PR #5, opened 2016-09-09
Directory clean-up

renekliment:
This brings:
- cleaner directory structure 
- deleted & ignored useless files

and it sets out base for another changes I've got coming.

@alexa-pi/coreteam (yay, we can use these team mentions now!) Please do a CR. There probably isn't anything that could go wrong, design-wise and implementation-wise, but I could have missed something - to paraphrase HIMYM: _Nothing good happens after midnight_. It's also a good exercise to stick to the rules I've mentioned - always check each other's code before merging.
Thank you!


maso27:
Great!
I'll look it over, and try a test build on my system as soon as I can. Hopefully tonight.

renekliment:
Awesome! To be honest, I haven't tested this, because I would have to do it on a Raspberry Pi still and I don't have mine ready yet.

maso27:
Alright!
I just installed it (my git-fu is weak, so I used _wget_ and _unzip_ to get the directory from here).
It installed smoothly and started right up!
It starts automatically after a reboot, and I chose to install shairport-sync, which works great too!

The only thing I noticed is that after the install, _vlc.py_ is in _scripts_ instead of _src_. I'm honestly surprised that it works like that.

renekliment:
Damn, I missed that. Should be fixed now. Thank you.
I'm gonna change this later anyway, but let's keep it always at a working state.

Please give it one more shot and if that's all, I'm gonna merge it and more interesting commits can come :-)
Also, I've finally set up my Raspberry Pi for testing my commits so I can avoid these embarrassing situations in the future - yay!

renekliment:
@maso27 Did you have any time to do a final check? I'd really like to merge this ASAP, so I can roll out the cool stuff. Thanks.

maso27:
Sorry, no.  I've got a fresh image on my pi now, and I'll give the update a run today.

(Also, is there a way to get just this commit through git?  I can clone branches, but couldn't get the SHA1 to point me to this build effectively.)

renekliment:
@maso27 Cool.

Sure. If you want a fresh directory to do that:
```
git clone -b directory-file-cleanup https://github.com/renekliment/AlexaPiNG.git
```

If you want to add it to your existing directory:
```
# Suppose you have done this already:
git clone https://github.com/alexa-pi/AlexaPi
cd AlexaPi
# Then do this
git remote add renekliment https://github.com/renekliment/AlexaPiNG.git
git fetch renekliment
git checkout renekliment/directory-file-cleanup
```

maso27:
The git notes worked great.  Thanks!

And the install went fine as well.  All works as it should.

As a note: The "always on monitoring" option is really tough to kill.  This was by design, but I'm wondering if we should add a line to _initd_alexa_monitored.sh_ under _stop)_ that'll kill _monitorAlexa.sh_ as well.  Otherwise, even _/etc/init.d/AlexaPi stop_ won't stop it.

Maybe I should do a pull request for that?

renekliment:
Awesome! Thanks. Merging then.

Yes, that would make sense. However, please wait a bit with that. I'm gonna do one quick PR regarding paths that's gonna change these files. Also, we need to refactor the init script a bit. Even though a systemd unit is gonna be used on most systems eventually, there should be a good init script for OpenWRT and such OSes. I've created an issue (#9) that sums up my thoughts on that. Feel free to take up any of the related subtasks.

Issue #9, opened 2016-09-14
Refactor init scripts


renekliment:
- [ ] write a systemd unit file that's gonna be used on the vast majority of platforms
- [ ] refactor the classic init scripts along with monitor
    - [ ] when issuing `/etc/init.d/AlexaPi stop`, the monitor should stop too
        - should we get rid of `/tmp/dont_start`?
    - [ ] merge `initd_alexa.sh` and `initd_alexa_monitored.sh`, because they're doing almost the same thing - use some configuration handling in the init script to find out if the monitor should be started too; ~~where do we store the config option?~~ ~~should use config from #11~~; setup script needs to reflect this too
    - [ ] use a `.pid` file which is unique - possibly `/tmp/AlexaPi.pid`
        - gets rid of `pgrep python -a | grep main.py` which could potentially hit another app
        - [ ] when the app starts a `.pid` file is written
        - [ ] the monitor checks that the file exists and a process with that PID exists, if not -> start an instance

renekliment:
@maso27 So if you want to take some of these things up, in #11, you can find a hint how the config should be handled. I'm not sure yet what the config variable (and section) should be called. 

I've got a question - why is the monitor actually there? Did you encounter crashes of AlexaPi? What was the reason? Or is it there _just in case_?

illperipherals:
Pulling a clean build still requires "root" to be edited to "opt" in two places. Both errors show up in the log. Verified at an Alexa Hackathon yesterday. Should this be rolled into an existing refactor issue or a new PR?

renekliment:
@illperipherals Are you pulling the `master` branch? I'm sorry for the misguided `README.md` in `master`, I don't know why I edited it before it the code was there.

There is currently a pending PR (#10) that gets rid of `/root` (or `/opt` for that matter) in the python code. You can clone it wherever you want. If you want AlexaPi to start on boot however, you have to clone it into `/opt`.

If you want to give it a try:
```
git clone -b universal-paths https://github.com/renekliment/AlexaPiNG.git AlexaPi
sudo ./AlexaPi/src/scripts/setup.sh
```  

Also, if you're a developer, familiar with the project and want to make it better, you're more than welcome to join in on the project. I've got a lot of ideas (and code) for it, but I'm in a desperate need for someone to review my PRs.

illperipherals:
[...]

Issue #4, opened 2016-09-09
[metabug] OS platforms


renekliment:
- [x] don't require root when not needed
- [x] install into `/opt`
- [x] universal paths generally (no `/home/pi` or `/root/`, ...)
- [x] [universal init files for all platforms](9)
- [x] [rewrite the setup script](60)

renekliment:
Last part of this fixed by #82.

PR #10, opened 2016-09-14
More sensible paths

renekliment:
This:
- installs to `/opt` now
- doesn't use `/root` anymore
- has some other path related things that seem more sensible now

This will eventually also lead to the app running under an ordinary user (not root).

I've considered storing the recordings/answers somewhere into `/var`, but decided to think that through thoroughly later.

@alexa-pi/coreteam Please test. Maybe @maso27 has some time to give it a try?


sammachin:
On the storing of recordings and answers piece there is a clause in the
Amazon T&C's that says you aren't supposed to store the voice files, it
seems like it's allowed to store temporary files in order to make the
device function but we should minimise this.
I'd suggest storing in /tmp so that they are cleared on a reboot. Also
prevents the storage filling up and causing issues.

sammachin:
I'll try installing on a fresh pi this weekend and report back

renekliment:
@sammachin Thanks for the clarification, I didn't know that. The code currently cleans up the directory on startup, but we can do better then. I thought about moving it out of `/opt`, since it's against FHS and what's more important, it bothers the file storage every time the user interacts with AlexaPi and it makes it harder to just copy the directory somewhere else / to someone else. So it's for sure we need to move it.

I did a little bit of reading on `/var` vs `/tmp` ([1](http://unix.stackexchange.com/questions/30489/what-is-the-difference-between-tmp-and-var-tmp), [2](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard), etc.) and you're right - it makes sense to store it in `/tmp`, which is often mounted on RAM and therefore:
* doesn't bother file storage like HDD, which can sleep instead if otherwise possible,
* doesn't take up space on the storage,
* cleans-up automatically on reboot.


My main concern when I thought about this before was that it takes up RAM (well, who would have guessed) and on memory constrained devices, like OpenWRT routers, where I want to run this too, it could be a problem. And like 512 MB RAM Pis are better off with memory not wasted. 
Thinking about it again:
* Having looked at the files, they're really small!
* Since a lot of the devices are gonna run 24/7 without a restart, we should implement periodical cleanups or a rule like _"store only last N files"_ so we don't reach some limit even with these small files.
* Depending on the rule, we should let the limit be configurable - for users with small RAM devices.
* If we decide on like "10 last answers", they're gonna take up ~300 kB and we also have the recording which should all fit into 0.5 MB and with this we can probably forget about the file size limit, because this is fine even for really small devices. I don't see a worst case scenario where this would go beyond 1 MB.
* So we can't use the files as cache? The file names look like they're unique for every request, even with the same answer, so I guess not :-(

I think moving it into `/tmp` and doing the _"last N files"_ is a good idea. If you agree, I'd file a separate issue for that and merge this as it is (if there are no other problems with this), since we need to implement the cleanup as well as creating the directory with proper permission, etc.

sammachin:
I'd have to check the specific details of the T&C's but I think the big one
is not storing users request speech files, Its mainly a privacy thing, I
don't think there's as much of an issue with the responses but we should
try to limit that anyway, I don't think we should cache responses if they
are duplicates just play back whatever AVS responds with.

My code always just had the last one as recording.wav partly for easy of
coding and partly to keep to the T&C's its useful to be able to do in and
replay your last command when trying to debug mic issues, users often had
silent requests as their mic wasn't setup right.

renekliment:
Ok, let's forget about the cache, it probably wouldn't be that useful anyway for 20K files and it might not even be possible.

With the `recording.wav` it still works the same, which I also think we should keep for debugging purposes. One last recording is fine IMHO.

So, do you agree that we'll do this in a separate issue? Which limiting method do you vote for (file size, number of files, combination, your own idea)? This is mainly to not to waste RAM.

sammachin:
Yep create an issue and we'll discuss :)

maso27:
Something worth being aware of, is that I put in a step to empty the tmpcontent directory every time Alexa finishes processing a command.  I didn't like the idea of that folder constantly filling with mp3's, and didn't know if Amazon liked it either.  The _--debug_ switch turns that emptying step off.

It could just as easily go into some subdirectory within /tmp (instantiated when Alexa initializes), and the "empty" step would just free up the RAM instead of HDD space.

renekliment:
@maso27 My mistake, thanks for the clarification! I have to admit, I just skimmed the code and thought that it cleans up only once per app load. I will read it now more carefully, since I'll be actually doing the code, not just changing paths, directory structure, etc. Sorry again for the confusion.
Issue solved then. I'll change the temp directory to be a subdir of /tmp along with its creation, permissions, etc. I'll send the commit to this PR first thing tomorrow morning. Please test then.

You're supposed to clone it into /opt as in the updated README in this PR. You have to still do it as root (or sudo from pi) and run it as root, but we're getting there for it to be able to run it as regular user.

maso27:
So I cloned it into _/opt_ and ran it as _pi_, with the _sudo_ command.  Everything worked great.

A question I've got, though, is what happens when someone tries cloning and installing it from a different directory, like _/root_?  At first glance, it appears you're storing _SCRIPT_DIRECTORY_ separate from _INSTALL_DIR_ in _setup.sh_.  

The script already does a check if you're running as root, can it also check if you're in the right directory?  Then it could either exit if it's wrong, or else perhaps copy the appropriate files into _/opt_ if you're starting from somewhere else?

I don't have time to check it now, but it seems that someone could do a janky install and never know it the way things are sitting. (Of course, they could do that before, but now I'm thinking about it a bit.)

renekliment:
So I originally wanted to just change the _tmp_path_ to `/tmp/something` and handle its creation/permissions/deletion, but then I remembered that I want for this to be cross-platform (even Windows eventually), so I went there for a multi-platform solution using python's _tempfile_ module. This way it should always go into `/tmp` if your machine is sane. If it is for some reason a little crazy, it should be handled in the best way possible.

renekliment:
@maso27 Thanks for pointing that out. So now the situation is that `SCRIPT_DIRECTORY` is where the install script is actually stored. The `INSTALL_DIR` is there just for pocketsphinx. 
Which brings me to the fact, that we need to think about pocketsphinx installation and also about hard-coding `/opt` or any other directory, because we want this cross-platform and for users to be able to run it from any other directory. Because if you don't care about about init scripts and want to test it manually the only problem is pocketsphinx path to the model.

So what do we do with pocketsphinx to
- enable users to use AlexaPi even if they don't download it to our _chosen_ path or don't have permissions to write there?
- have the code cross-platform?
- reduce its size? The current `git clone` takes ~200 MB.
    - If we add `--depth=1` to the `clone`, it takes _only_ ~30 MB.

I see these options:
- we could have pocketsphinx as a subdirectory of AlexaPi, 
    - cloning it in the setup script, 
    - or even having it as a git submodule
- we could even distribute the model only (if that's all that's needed and the license allows this) directly in our repo as a subdirectory

I'm tired so I can't tell which option sounds best.
Let me know what you think. Thank you.

maso27:
Yeah, that git clone of _pocketsphinx_ bugs me. It's available through _apt-get_, but I didn't have luck making that version work. In setup.sh, it's installed with _pip_, but I never found a directory to point to that made things happy either. So I just clone it and point main.py to it.

It's possible that the dictionary file (and realistically, just the _Alexa = "ah leks ah"_ portion of that) is all that's required. But I have a sneaking suspicion it involves more than that. Might be possible to really pare that down though.

But I like the idea of moving whatever we need into the AlexaPi directory. I made it separate.back when it was building and installing a number of different programs and I wanted to keep track. That can be changed now.

Incidentally, a number of distros have shairport-sync in their apt repos. For some.reason it isn't in Raspbian so I build it from source.

renekliment:
@maso27 Yep, I've looked at those packages in repo and they didn't seem complete.

I am for keeping the whole dictionary, because, correct me if I'm wrong, users can set their own trigger word now.

What I'm planning to do regarding #4 is to make the setup script modular and depending on the OS (or even Linux distribution) if would first try to install stuff from the system repos and then fall back to building it from source or whatever. This would cover both _pocketsphinx_ and _shairport-sync_.

What I'm thinking now (and just saw this over at some other repo) is to create a `thirdparty` directory inside `AlexaPi/src` and download/clone stuff in there when needed (when it defaults to that). If there was a working X installed by a system method, we could just symlink `thirdparty/X` to that.

Thinking about this, I actually quite like it. What do you think?

renekliment:
So I looked at the PyPI version of _pocketsphinx_ and I was able to make it work with it. So this solves all of our issues with _pocketsphinx_ install (
- huge repo clone wasting storage space, internet data and time,
- wanting to have it multi-platform (even Windows and OS X)
- and getting rid of hard-coded paths to run it

).

**MIC. DROP!**

renekliment:
So this should be all from me regarding this PR. I'd like to merge this and move on to #11 when you test it and if you don't have any other issues with it. Thanks!

maso27:
@renekliment 
Glad to hear it on the _pocketsphinx_ front!
I'm working on a fresh install to test.  Will report back later.

maso27:
@renekliment 
Oh, to answer your question about the always-on monitoring: I developed that when I was trying to port to an Arch Linux box.  My install was really crashy and I came up with it as a means to keep it running anyway.

I haven't needed it on any raspbian configs, ever.  As far as I'm concerned it can go away.

renekliment:
@maso27 Awesome. I'm also excited about this elegant solution :-) Let me know when you've tested that.

I see. I think we should keep the monitor, but merge the init script with the other one as indicated in #9. On most systems, systemd will handle that, but we need to write the unit file first. I might actually have a look at that, since it doesn't collide with any of the other architectural changes (other than this PR so we have the path in `/opt`).

maso27:
So my fresh install went well.  I'm not at home, so I can't talk to her until this evening, but Alexa runs and reports "connection OK".  Starts again successfully after a reboot.

I tried to install from _/home/pi_ first, just to see what would happen.  It alarmed and let me terminate as expected.

I've installed _git_ and _screen_ in addition to the base install (with _shairport-sync_ enabled).  And after everything is installed, it takes up 1.6 GB of space.  So it would even fit on my dinky 2 GB SD cards!

Good work.

renekliment:
Okay, I'm a terrible liar. There's one more commit to this PR from me. But this should be it - seriously!

Also, I've got some sweet init scripts refactor on my local machine waiting - say hello to systemd unit file, running it under an ordinary user, pid files, etc. Yay!

sammachin:
Ok I've finally had a chance to test this, first off this is what I did starting from a fresh install of raspbian 2016-05-27.
```
ssh pi@raspberrypi.local
sudo bash
cd /opt
git clone https://github.com/renekliment/AlexaPiNG.git
cd AlexaPiNG/src/scripts
./setup
```

I went with the no shairport and no monitoring option.

Install went OK apart from the AlexaPi init.d script has the old path to /root/ for startup so it doesn't auto start. After I corrected that it runs fine.

I've noticed a couple of differences from my version I'm guessing they came in in @maso27's changes:
1) If you are running in non wakeword mode then the button is now a press & release to talk rather than a press and hold to talk. I guess this means we are using some for of speech detection? This means there is a slight lag before it starts listening. Also in some use cases I have found the Press and Hold to talk to be a better experience especially in noisy environments or for people who's speech isn't great, you can add longer pauses. I can see the flip side that press and release is also nice and more like the way the Amazon TAP works. 

The original version had a 3rd LED state "processing" essentially this was both LED's on while the request was being send to amazon and waiting for a response, if using a dual Red/Green LED this gave me a yellow light. So the indication was Red (Recording) > Yellow (Processing) > Green (Playing)
We now have Red LED (recording) > LED off for a few secs > Green LED (playing)

@renekliment I think if you can fix that startup script issue then I can work on putting back those previous features in a new PR later.

renekliment:
@sammachin Thank you for testing. You have however used my _master_ branch, which is even with the official _master _ as it should be. You need to do `git checkout universal-paths` after the `git clone`. Which is why the path was incorrect, because it was using the old code, not from this PR.

ad 1: Yes, there is a speech detection from maso27 so you can speak as long as you want (there is no timer). The way it works now is that a button triggers Alexa and optionally while you are holding the button it forces her to listen, which you can use if you need to make a pause while you're thinking what to say next.

ad LEDs: Yeah, I've noticed it's little a different in the code, but I don't use LEDs (I'll give it a try later) so I haven't noticed what. However, please wait with this until device_platforms get merged, because I changed the code regarding this a lot in my local branch. You can file an issue in the meantime.

sammachin:
ah yeah I cloned the repo and forgot to switch branches!

Just about to test on another pi using the detection option will use the
correct branch this time :)

Ah thats nice so we have both tap and hold options.

maso27:
@sammachin
Yep, you've seen some of the things that changed between my version1.1 and version1.2 branches.  The silence detection for voice did change the way the button functions, in that a tap will begin a listening process in addition to it continuing to record as long as the button is held.

The LED functionality was changed to provide distinct feedback to the user for when the box is listening. Having both LEDs illuminate is misleading when only one has been implemented, or when they are in separate locations.
You can see it for yourself on your own install, but here's how I've been using it:
https://youtu.be/wlj4N7HQYs4
(You can see how the listening LED works at 2:20.)

And it's fine if you want to change how they're working. I just wanted to illustrate why I did what I did.

sammachin:
@maso27 
Yeah fair points I've always used dual color LED so I've been expecting those indications, 
The ones in my latest hardware are RGB LED's with 3 pins so we could also move to using a 3rd pin for monitoring, personally i do like having some kind of visual feedback when its completed the record and is working so you know 'something' is happening.

One of my eventual goals would be to make the code modular enough that people can use anything they like for indication and we just call some sort of 'event' to say the state has changed, then you can hookup whatever LED's other devices you like, even using something like smart light bulbs across the network!

Only thing I've found with the silence detection now is that there is a slight ~750ms delay between pressing the button and LED coming on to start recording. I'll look into that a bit more.

sammachin:
@renekliment just run 2 more installs again with a fresh raspbian;
```
ssh pi@raspberrypi.local
sudo bash
cd /opt
git clone -b universal-paths https://github.com/renekliment/AlexaPiNG.git
mv AlexaPiNG AlexaPi
cd AlexaPi/src/scripts
./setup.sh
```

First one was done with no shairport no monitoring, all ok this time with the startup script.
However, even though I've selected no monitoring she is still responding to the Alexa wakeword!

Second one was with no shairport but monitoring yes.
This works as expected and detects the wakeword most of the time.

maso27:
The "always-on monitoring" option isn't to disable the wakeword. 
It instantiates a script that runs behind the scenes and turns AlexaPi back on any time it stops. It turns on in _silent_ mode so it doesn't say "hello" and startle the whole family when it happens.
It was designed when I was working on a crashy arch Linux port, but really isn't necessary.

renekliment:
@sammachin Actually the monitoring is the monitoring of AlexaPi process so it gets started again when it crashes. I know it's confusing. I've changed in in my local branch regarding refactoring of init scripts.

sammachin:
ah right ok that makes sense I thought we had an option to choose if you wanted wakeword/monitoring.

So the only feature choice is to include shairport then?

renekliment:
Yes.

renekliment:
So what's your opinion on this PR @alexa-pi/coreteam? Ready to be merged?

maso27:
I just did a reinstall with your updates from two days ago.
No issues.
So yeah, as far as I'm concerned it's ready.

renekliment:
Great! I can't wait for the things that can come now. Merging.

PR #11, opened 2016-09-17
[WIP] Common YAML config

renekliment:
So this is what I would like it to be. 

It is based on #10, therefore #10 needs to be merged first and this need to be rebased. I've decided to publish this as it is, so we can discuss it ASAP.

@alexa-pi/coreteam Let me know what you think.


renekliment:
So, with what I've written in #14, can we have the trigger phrase in the config file? I mean the users could have previously changed it in the code as well. This however could be considered as giving them a real option and encouraging them to do it. There's practically nothing we can do to take away the option from them when using pocketsphinx (aside from using a model with _that one word_ only, which someone would have to do and I'm not that eager to do that). If I'm correct when using `snowboy`, this will be not an issue since there would be only a model for _the one word_.

illperipherals:
snowboy will allow multiple words in a model.

renekliment:
Superseded by #18.

PR #15, opened 2016-09-21
Init scripts refactor

renekliment:
This fixes all points in #9.

Apart from those points, it:
- sets systemd loading as default,
- runs _AlexaPi_ under a non-privileged user _alexapi_. - Yes, finally! 

Important remarks:
- You **need** to be on Raspbian Jessie, because it runs under a normal user and touches GPIO which wasn't possible on older versions. I hope folks at Arch Linux implemented the same thing, since it should be in kernel anyway. @maso27 Can you test please?
- When systemd restarts _AlexaPi_ on crash, it isn't run with the _silent_ option, therefore says _Hello_ again. I wasn't able to find a way to achieve this. But I don't think it's such a problem. That being also due to the fact, that these watchdogs are borderline cases and AFAIK there is no need for it. It's there just in case.
- If you're gonna try the classic init and then native systemd, apart from `rm /etc/init.d/AlexaPi` and installing with the new option, you have to actually restart the Pi, in order for systemd to reload the change for some reason. I guess it's confused because it previously had to use a non-native file and now we want to use native ... I don't know, I doesn't really matter.
- It won't work on OpenWRT, because the `start-stop-daemon` there doesn't offer all the capabilities a full-sized does. We'll probably have to have a special init script for that OS. It's probably gonna be like the one that was here before, but I didn't want to keep it alongside, because we need to test anyway, so when we get there (_AlexaPi_ running even on OpenWRT), we'll make one.

It is also based on #10, therefore #10 needs to be merged first and this need to be rebased. I've decided to publish this as it is, so we can discuss it ASAP.


renekliment:
I'm terribly sorry for the commit history. I originally wanted to just merge #10, but then I decided to squash it, because it made more sense and now we've got this. I'm gonna be much more careful in the future. However, I'm gonna squash this too, so it doesn't really matter now.

sammachin:
Does this one need testing or is it superseded by #18 ?

renekliment:
Yes, it needs testing. #18 is about config, this is something different. I'm gonna rebase this so it has clean git history and no conflicts. I'll get back with a new PR.

PR #16, opened 2016-09-21
Basic Travis CI setup

renekliment:
Travis CI now displays a nasty message that the checks have failed because it tries to run ruby tests on our code.
We don't have tests yet, but we can at least set it to python and check our source code with pylint (syntax error, style errors, recommendations, ...). I've set it to errors only now, because there is a lot (and I mean really a lot) of things we need to address before we can let it display more than that.

The checks are gonna fail for a while now before I get to pushing some commits that fix these issues (they're non critical). We need to merge the refactoring stuff that's ready first - configuration, init scripts, device_platforms, etc.

Anyways, that's it for now. I'd like to merge this as it is, even when it doesn't change much really. Is that okay @alexa-pi/coreteam? Thanks.


sammachin:
I don't see an issue with merging, this whole repo/project is WIP anyway so its not an issue if we have error messages so long as we know to ignore them

renekliment:
@sammachin Great. The point of this is to at least have **python** code errors that you can look at, not silly errors like "you don't have some **ruby** stuff" - no shit.
Updated due to new config now in master and changed some params so we have more accurate pylint error messages.
The rationale is still the same, therefore I'm merging this.

PR #17, opened 2016-09-21
Common config

renekliment:
So this is my idea of the configuration done right :-) It's gonna change a bit with my device_platforms PR, but this is it for now.

This PR replaces #11 as it was a mess when #10 got pulled in.


renekliment:
Damn, branched off a wrong branch :-/

PR #8, opened 2016-09-10
CONTRIBUTING.md - written in details

renekliment:
Update so everyone knows what to do when they wanna send some code.

PR #18, opened 2016-09-21
Configuration refactor

renekliment:
So this is my idea of the configuration done right :-) It's gonna change a bit with my device_platforms PR, but this is it for now.

This PR replaces #11 as it was a mess when #10 got pulled in.

The third time's a charmer!


maso27:
I'm trying this, and I think I've seen a couple of issues.
I can't go back and re-verify this until later (I'm tunnelling into the pi remotely), but here's what happened:

 -Ran `sudo ./AlexaPi/src/scripts/setup.sh` from opt, as listed in the readme.  I had been running from within  _scripts_ previously but read this in the instructions and wanted to check it out.

 -Install seemed to go well, but I bumped the enter key when generating _config.yaml_.  So I knew it was screwed up.  Finished anyway.

 -Re-ran setup.sh so that it would allow me to re-generate _config.yaml_.  It told me that it already existed, and that it would use the existing one, but **proceeded to ask for the details anyway**.  I entered them and they were integrated into _config.yaml_.

 -Rebooted.  `ps ax | grep main.py` did not show any processes.

 -Ran `python /opt/AlexaPi/src/main.py`. Received:
`  File "/opt/AlexaPi/src/main.py", line 36, in <module>
    with open(alexapi.config.filename, 'r') as stream:
IOError: [Errno 2] No such file or directory: 'config.yaml'
`

So it seems maybe some attention should be paid to how _setup.sh_ handles a pre-existing _config.yaml_ file, and maybe _main.py_ is looking for _config.yaml_ in the wrong place.

renekliment:
@maso27 Thanks for testing!

- You can actually run the setup script from anywhere, since it `cd`s in the directories correctly. The README was wrong before and I've decided for this variant.
- config setup: Yeah, I've been thinking about that one, but haven't decided what the behavior should be. It's confusing what it says now, okay. The message means, that it won't overwrite the whole file with a new one, therefore destroying all your other changes. It means that it changes only the variables it asks you for. 
I've been thinking that we could load the current variables from the file and offer them to the user as default (on Enter key). Or we could probably just end with the message that the config file exists? I'm not sure, what do you think?
- config path: Oh, yeah, definitely a bug. I've probably only run _AlexaPi_ with this from the same directory and it didn't hit me. Will fix that.

maso27:
- Yeah, I figured you had the directories all buttoned down, but just wanted to try from a different angle anyway.
- On the config: this may or may not be easy, but what about reading in the values from the existing _config.yaml_ and presenting them to the user, with a prompt to keep them?
Something like:
```
Your current ProductID: RaspberryEcho
Keep (Y/n)?
```
Or else a prompt for whether the user wants to keep the existing _config.yaml_ settings?

I don't know.  Just ideas.

renekliment:
Ok, so how about this:
- no config -> as usual
- existing config -> options
  - [0] keep (this would exit the script as there's nothing else to do)
  - [1] edit existing (this would use the existing, but look like no config and would edit only the creds)
    ```
Editing existing configuration ...
Hit Enter to fill in the current value (in brackets).

Enter your Device Type ID ... [MyAlexaPi]:
```
  - [2] start with a fresh one (this would delete the old one and behave as with no config)

maso27:
Yeah, that makes sense to me.

maso27:
So...
Should I be testing any particular pull requests right now?  I'm happy to check something out, but it seems like a number of them have been superseded and I wasn't sure what would be best to focus on.

renekliment:
@maso27 Please have a look at #8 and #19. That's just documentation update.
You can review #15, that's another major refactor.

I'm currently working on fixing this one - I need to rethink some things a bit.

renekliment:
@alexa-pi/coreteam Okay, I gave this a bit more thought and here it goes. The mentioned bugs should be fixed. Please test.

renekliment:
@maso27 P.S.: All open PRs are pending review / testing / approval, if there is not `[WIP]` in the name. 
There were 2 attempts (PRs) on the config refactor that I had to close due to messed up git history. Hence the name of this branch :-)

maso27:
Looks great!
I ran setup.sh through a few times, first checking if it kept the old config file, then seeing if I could edit one line, then seeing if I could start a new one.

Everything worked!

renekliment:
@maso27 Great! Let me know when you think this is merge-ready. Thanks.

sammachin:
Tested this morning on a Pi2 and Pi3 seems good to me, 
OK To merge

renekliment:
Awesome, merged!

PR #19, opened 2016-09-21
Docs update: detailed instructions for Amazon device registration

renekliment:
What the title says.

@alexa-pi/coreteam Is this ok?


sammachin:
When it comes to the security profile URL's I've been using http://raspberrypi.local instead of the IP address. Raspbian has MDNS support so you can just use the default hostname and it doesn't matter what IP address DHCP gives it.

I've been changing all my alexapi's hostnames to alexapi so its now alexapi.local might be something we could do in the install script?

illperipherals:
Mason, have you looked at the Pi Zero install using Bonjour for .local?

renekliment:
@sammachin So I've googled mDNS and it looks like a pretty dope tech! Just tried it now on my laptop.
Do you know what's the support across operating systems? I think OS X is fine, because it should have Bonjour enabled by default. I googled a bit about Windows and it doesn't seem that great - it should be there in Windows 10 or you can install something. Linux experiences may vary I guess - I had to install `nss-mdns` package on my Arch.

maso27:
I've got no pi zero, but I've got a C.H.I.P. computer I can try out when I get the spare time to set it up.

As far as .local is concerned, it's used on the config for my 3D printer controller, and I've had spotty success with it.  I didn't go to far into it-might've been an overly restrictive work network that caused issues.

I liked the idea for setting an alexapi hostname until I thought about other use cases. If someone installs this on a multi-use system and it changes their hostname, they might be miffed. Perhaps another option in config?

renekliment:
We definitely wouldn't change the hostname, but we could advertise `alexapi.local` as additional one. Or we could just stick to whatever the system uses, if it's there (probably `raspberrypi.local` on RPis) if the former isn't possible.

sammachin:
@illperipherals THe Pi Zero should be exactly the same for mDNS as its still running Raspbian.

I'll check the CHIP later, got to set one up for testing anyway.

@renekliment good point on checking windows I think this needs more investigation.

The docs looks good though so I think we should go ahead and merge this then I'll raise an issue about mDNS.

renekliment:
@sammachin Great. I agree - this is it for now and if we can use mDNS, we'll add support later after we investigate. Feel free to open an issue. Merging!

PR #28, opened 2016-09-25
Init scripts refactor - fixes #9

### LINKED ISSUE #9: Refactor init scripts

renekliment:
This supersedes #15, is now rebased on master.

This fixes all points in #9.
Apart from those points, it:
- sets systemd loading as default,
- runs _AlexaPi_ under a non-privileged user _alexapi_. - Yes, finally! 

Important remarks:
- You **need** to be on Raspbian Jessie, because it runs under a normal user and touches GPIO which wasn't possible on older versions. I hope folks at Arch Linux implemented the same thing, since it should be in kernel anyway. @maso27 Can you test please?
- When systemd restarts _AlexaPi_ on crash, it isn't run with the _silent_ option, therefore says _Hello_ again. I wasn't able to find a way to achieve this. But I don't think it's such a problem. That being also due to the fact, that these watchdogs are borderline cases and AFAIK there is no need for it. It's there just in case.
- If you're gonna try the classic init and then native systemd, apart from `rm /etc/init.d/AlexaPi` and installing with the new option, you have to actually restart the Pi, in order for systemd to reload the change for some reason. I guess it's confused because it previously had to use a non-native file and now we want to use native ... I don't know, I doesn't really matter.
- It won't work on OpenWRT, because the `start-stop-daemon` there doesn't offer all the capabilities a full-sized does. We'll probably have to have a special init script for that OS. It's probably gonna be like the one that was here before, but I didn't want to keep it alongside, because we need to test anyway, so when we get there (_AlexaPi_ running even on OpenWRT), we'll make one.


maso27:
I installed this today (in raspbian) and exercised the systemctl commands.  Works well.

I'll look into getting it into Arch next.  I haven't played with the pi's GPIO in Arch: I'll have to see how it works.

renekliment:
Great. You'll have to do the install manually though for now.

renekliment:
@alexa-pi/coreteam Any progress on CR / testing?

maso27:
I'm having trouble getting arch onto a pi.  I was working on a pi 1b, but that image from archlinuxarm isn't working for me.

I need some time to try the pi 2 image on my pi 2 board (my son wanted some music and has "borrowed" it with the most recent Alexa install).

maso27:
Ok.  It took a little time, but I got it to run on Arch Linux on the Pi 2. (Had to create the image from another Arch box, for some reason it never worked under Raspbian.)

In case folks are interested, here's what I had to do:
- "python" => "python2" everywhere (including all other scripts)
- "apt-get update" => "pacman -Sy"
- "apt-get install xxxx -y" => "pacman -S xxxx --noconfirm"
- "pip" => "pip2"
- "python-dev" => "python2"
- "libasound2-dev" => "alsa-lib" and "alsa-utils"
- "libpulse-dev" => "libpulse"
- remove "python-alsaaudio"
- add "gcc" to list of things to install
- add "RPi.GPIO=>0.6.2" to requirements.txt
- add "dtparam=audio=on" to /boot/config.txt
- _shairport-sync_ is already a package in archlinux repos, so just "pacman -S shairport-sync --noconfirm" to replace most of the build steps there.

I ran the install as root, and systemd doesn't seem to be working with user alexapi.
I changed _user_ and _group_ to root in AlexaPi.service, and it starts at boot.

Not sure that any of this is what you had in mind, but it's cool to have it working.

I'd say, from a raspbian perspective, that this is ready to merge.
I'm not sure how we would handle multiple architectures though.

renekliment:
@maso27 Thank you for the list - we'll use that when creating OS platforms to support Arch!

I wanted someone to test (/ getting it to work) running in under the _alexapi_ user to see if that also works by default in Arch. Anyway, I have a spare SD card, so I'm gonna give it a shot.

illperipherals:
@maso27 Nice work, man! I am testing inits-refactor-2 with Arch on a Pi3 and a headless Zero right now. Did you happen to look at "trick the environment" at https://wiki.archlinux.org/index.php/Python ?

renekliment:
Alright. I can see that it's not there by default. We need to add the group _gpio_, if it's not there, add the user _alexapi_ there and set some _udev_ rules - [[1]](https://sourceforge.net/p/raspberry-gpio-python/tickets/115/), [[2]](https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=9667#p734582). @maso27 @illperipherals You can try this if you want to and see if it works for you so we can do it. Running under non-root is an important thing.

However I'd do that in a separate PR. Maybe along with the OS platforms so it all sets up automatically?
It's a small change, but seems to be quite off topic for this PR and it will require additional testing. And we don't advertise that we support other distributions already, since it's not finished yet (even though even I personally have it running on Arch after manual install).

So OK to merge this as it is?

illperipherals:
I am wrapping up my config on the Zero right now. One small thing that has bitten me twice is when you get to the part on whether you want to retain your config file... if you hit the enter key anywhere in the install process by accident, it feeds you through that...

Not a huge deal, but thought wrapping with stty might fit under this PR?
Thoughts?

renekliment:
@illperipherals Can you elaborate or post a screenshot? I'm not entirely sure what you mean.

illperipherals:
@renekliment When asked for the Amazon credentials in setup.sh, it allows empty values (from hitting CR during the long foreground process).

illperipherals:
@maso27 What does your system load look like on this build?

renekliment:
@illperipherals I see. Please file an issue regarding that, since it doesn't concern init scripts / starting the app and should therefore come in another PR.

maso27:
@illperipherals I hadn't seen the "trick the environment" thing, but that could prove quite useful.  I use OctoPrint with my 3D printers, and they recently switched to a virtual environment for running it.  Worth investigation, for sure.

And on the Arch install with an RPi2, it looks like the load of a quiescent alexa is 2.6% CPU, 5.4% memory.
Better than I expected.

@renekliment I definitely agree that it's good to get away from the root requirement.  I'll look into adding the GPIO group and device when next I get the chance.

But yeah, I'd say this is good to merge.

renekliment:
Great, merging.

ad Arch for RPi: It's a shame that it's not there by default as in Raspbian, but whatever. Arch is kinda DIY and we can totally do this by ourselves.

maso27:
I know it's off-topic, but I got Arch to start with user/group alexapi. 

I figured since I started writing about it here I'd put what I had so others could see it.

My steps:
`groupadd gpio -r`
`usermod -a -G gpio alexapi`
`nano /etc/udev/rules.d/99-gpio.rules`
Insert the following (3 lines):
`SUBSYSTEM=="bcm2835-gpiomem", KERNEL=="gpiomem", GROUP="gpio", MODE="0660"
SUBSYSTEM=="gpio", KERNEL=="gpiochip*", ACTION=="add", PROGRAM="/bin/sh -c 'chown root:gpio /sys/class/gpio/export /sys/class/gpio/unexport ; chmod 220 /sys/class/gpio/export /sys/class/gpio/unexport'"
SUBSYSTEM=="gpio", KERNEL=="gpio*", ACTION=="add", PROGRAM="/bin/sh -c 'chown root:gpio /sys%p/active_low /sys%p/direction /sys%p/edge /sys%p/value ; chmod 660 /sys%p/active_low /sys%p/direction /sys%p/edge /sys%p/value'"`

Reboot, and voila!

renekliment:
@maso27 Awesome! We'll somehow include this, so users don't have to do it themselves.

Issue #24, opened 2016-09-24
Refactor: triggering


renekliment:
Currently, it's kinda messy. We need to make it a little more general / modular.

We definitely need to support:
- device_platform triggering
- pocketsphinx / snowboy triggering
- keypress triggering when not running in daemon mode

sammachin:
what do you mean by `device_platform` triggering.

I'm keen to bring back the option of 'press to talk' only mode like in my original, I was giving a talk the other eve with a Pi running this codebase on the desk and the amount of times it triggered when I said "Alexa" was crazy!

Been thinking about adding support for Keyboard/HID triggers for a while, I've got one of the RasPi SenseHats which has a grid of neopixels and a joystick which mimicks keyboard inputs would be nice to have that as a hardware option for the Pi as thats then soldering free for users.

renekliment:
@sammachin I'm sorry I keep mentioning the _device_platforms_ thing when there's no public code. It's exactly what you mean by #26. I have some code ready on my local machine, but needed to merge the config first. I'll rebase and publish it ASAP.

Yes, we'll definitely make the voice activation configurable.

We can then rework the the things you mention once we have #26.

sammachin:
Ok would be good to see what you're doing ASAP I've got a little time today to do some coding (finally!) and I was gonna work on the indications as I'm dying to bring back the processing state LED its really starting to bug me not having that.

renekliment:
Damn, forgot to close by commit message - again! Fixed by #119.

Issue #6, opened 2016-09-10
Merge work in forks


renekliment:
There are various forks of the original project. Moreover, there are forks of those forks. The goal of this is that when a common code base is provided, we can unify the work of so many people, so we can work together to be effective, do not duplicate work and not to confuse end users.

Known forks that we would like to merge eventually (actually rather soon):
- [ ] @flooie https://github.com/flooie/AlexaPi/
- [ ] @sammachin https://github.com/sammachin/AlexaCHIP
    - [ ] @jitto https://github.com/jitto/AlexaCHIP
    - [ ] @johnwskinner https://github.com/johnwskinner/AlexaCHIP
    - [ ] @billzajac https://github.com/billzajac/AlexaCHIP
- [ ] @dony71 https://github.com/dony71/AlexaOPi
    - [ ] @GadgetReactor https://github.com/GadgetReactor/AlexaOPi

If you are one of the devs mentioned above and seeing this, please consider joining forces with us. The new code base that is gonna support multiple devices and OS platforms is gonna be ready in a week or two for testing and implementing new things. This is a note so you might consider this. If you have any questions, do not hesitate to ask in comments. Thank you!

flooie:
@renekliment happy to help.  just let me know what you need me to do.  I'm not a professional dev.

jitto:
@renekliment My changes were to run in python3 and use USB audio and a few months old.  If these are not already in upstream, I can definitely create a PR once the new code base is ready.

renekliment:
Awesome, glad to hear that :-) I'll let you know once the new code base is ready.

PhyberApex:
Reporting in. I don't know what the "current" code base is but I did the voice activation by a third party thing called "Snowboy" so I am ready to help out once the codebase is up.

renekliment:
Also, if any of you has some time and wants to at least review Pull Requests in this project, please contact @sammachin and ask him to add you to the organization. Thanks!

renekliment:
So before we're ready to merge the changes in your forks, we have to first stabilize the new code base. However, it seems that there's not so many people with much free time to help out with that. Do you guys (@flooie, @PhyberApex, @jitto, others mentioned) have any free time to do code review / testing of the PRs I'm pushing? 

Sorry for spamming so much and thanks!

flooie:
@renekliment - Unfortunately, I am starting a new job tomorrow - so my availability will be significantly limited for the foreseeable future.

[...]

PR #30, opened 2016-09-25
Device platforms - fixes #26

### LINKED ISSUE #26: Refactor: Device platforms
I'd like to abstract the notifications out from the main code base into their own functions that then call whatever hardware we might be using, so I propose a number of functions like `ind_record_on` `ind_record_off` `ind_ready` etc.

renekliment:
So here it goes. So far it's quite a silly code.

Two things:
- the triggering mechanism probably needs to get some rework - callbacks and stuff instead of polling the `should_record()`
- the stuff in the _desktop_ platform should be common for all _AlexaPi_ platforms and should not require `[Enter]`

Both things should be however addressed in #24, because this PR solves only the abstraction / making it modular. It doesn't and shouldn't change the logic. 

@sammachin So this is how I'd do it. If / when this gets merged, feel free to do the thing you wanna do. I suggest adding a `platform.indicate_processing(state=True)` method. But we can discuss that in the corresponding issue / PR.


renekliment:
@alexa-pi/coreteam Any progress on CR / testing?

renekliment:
So I gave this another thought and we could also look at it as a machine with states:
- ready
- processing
- listening
- recording

and we'd implement it like `platform.switch_state(platform.READY)` or `set_state` or whatever. This would be better, because now if we want to go from one state to another, we have to call two functions - to de-indicate one state to indicate another one.

The question is, if these states are truly exclusive and therefore make sense.

sammachin:
To me there are 2 things you may want to indicate:

States : idle/recording/processing/playing
Events : ready/error(s)
The machine would always be in one of the above states but I think events
would only occur in the idle state?

maso27:
I can't think of anywhere those states step on each other, but there are a couple of events that deviate from the main "waiting for input" routine.

- Button is pressed for 10 seconds to initiate shutdown.  This happens from within a "recording" state because the button press has triggered recording as well.
- Secondary input is required.  This happens when Alexa asks for more information, and we jump from "playing" straight to "recording."

(Admittedly, I haven't had time to look at this pull request, so disregard if that's not useful information.)

renekliment:
@sammachin When would _ready_ happen? Once after successful startup?
Regarding errors - how about also adding a state called `failed`?
I believe events could happen in any state - for example when the audio device can't be accessed in playing state, AVS returns an error when processing or something like that.

So you agree to having this rewritten from `indicate_something(blah)` to `set_state(blah)`?

maso27:
I'm trying this out, and here's what I've found:

1. The button press triggers Alexa to say "yes" before she starts listening. That's different, and probably not necessary.  The reason I added the "yes" in the first place was to provide feedback that the trigger word had been detected.

2. The ten-second button-hold to shut down is not there. I personally use this when I move Alexa from one room to another.

3. I hadn't checked this in a while, but "news flash" doesn't work on this version. I happened to check it on my kitchen installation (running my version 1.2) this morning and everything played okay. On this one, everything plays on top of each other and pressing the button doesn't stop it.
(Also worth noting though, is that in my 1.2 version you never hear Alexa's voice either. In this one she says "news flash...." and "weather" simultaneously before spewing a wall of sound.)

renekliment:
@maso27 Thank you for the feedback.

1. That's on purpose. I wanted it that way. But you're right some people don't. My idea behind this was that I want feedback, to make sure it listens (the button does something) and that people might connect some other things than buttons (it's just a matter of connecting two electrical leads). Let's make it configurable then.
2. Yes, I've commented that out for the time being. Seeing it's really being used, I'd go for a configurable command in the _raspberrypi_ platform. So people can disable it / set some other action. What do you say?
3. That very probably isn't caused by this PR, is in _master_ and therefore is an issue that should be fixed separately. I believe @davidmroth fixed the overlapping in #42.

@alexa-pi/coreteam What do you think about rewriting this to _states_ as mentioned above? Or should we merge this as it is and rewrite it later so we move forward? Please let me know.

renekliment:
@ajplante I see you've been doing some stuff at https://github.com/ajplante/AlexaPi-1/tree/h3 
Please base it on this PR. We want to have one repo for all the platforms, therefore we're making this modular. Also, please stick with the rest of the principles in this repo - keep using the configuration, etc.

P.S.: I've got an OrangePi I want to run this on, so I'm excited about that :) I'm also interested in the comparison of the py library vs. sysfs approach.

maso27:
@renekliment
Yeah, I agree that maybe the "yes" could be configurable.
I think there are a couple other points worth discussing though.

I believe this breaks the part where Alexa keeps listening as long as the button is pressed. If the user pauses while making a query (as my kids do), the silence listener can time out. They may just need to adjust their habits, but it's different and worth noting.

Second, the button is now quite sensitive. I've had mine trigger multiple times through the static from my finger before I press the button. Is there a way to de-bounce it a bit?

And I'd like to look back at when "news flash" broke. Since this was based off my version 1.2, it was working when we started. Any idea when it might've happened?

renekliment:
> I believe this breaks the part where Alexa keeps listening as long as the button is pressed.

What do you mean by _this_? The _long button press_? That's the main reason I wanted to have it disabled. I'll make it configurable and then people can decide what they habits are and set it / use it accordingly. I don't see any other options with this other than adding another button or doing some morse code stuff (just joking) Another potential candidate for a configuration option could be the press time.

That's weird. I haven't changed the debouncing settings.

~~Okay, that's also weird.~~ I might have an idea! asmagill discovered and fixed #34, so we actually weren't playing one audio file before. And now that we play them all, they overlap, because we don't do it in serial (I'm not sure if that's proper English). This should however be fixed in #42. Does that make sense, or are you experiencing something different?

ajplante:
@renekliment no problem. I have just been trying to get the GPIO libraries to interface properly. Do you mean you want one branch for all the platforms? I am guessing we are trying to get the setup script to detect the platform and use the correct values for it.

renekliment:
@ajplante Yes, exactly :) We'll have one version instead of the mess that is right now. You can checkout the AlexaOPi repos we've been tracking - #6. This is gonna benefit both end-users and developers.

We haven't had time to refactor the setup script yet. We'll get there eventually, but now, as developers, it's fine to do necessary steps manually. We need to make AlexaPi work first :)

Anyways, play as you like, but when you want to merge this back, please have it ready for the modular base we're working on.

Thanks!

maso27:
@renekliment 
Yeah, considering you don't hear Alexa's voice at all during **"news flash"** on version 1.2, I'd bet that some fix to get that to work broke the serial playback that _was_ working.

On the **long button press**, there are two functionalities that have broken as a result of the button trigger now being a single event, instead of a threaded process.

- OLD: would record as long as the button is pressed, regardless of whether silence had been detected.
- NEW: button press begins silence listening process.

- OLD: if button had been pressed for a total of 10 seconds, system shutdown was initiated.
- NEW: not present.
This one might not be possible anymore, since we're running as user _alexapi_ instead of _root_, but it was nice to have, as someone who doesn't like pulling the plug on a running linux system.

As for the **debounce** issue, in version 1.2, the threaded event had an inherent debounce value, set as _bouncetime=100_ when it was called.
Where does the button detection happen in this code?  I haven't encountered _platform.should_record()_ before, so forgive my ignorance.

Also, I don't know if I've said it enough, but thank you so much for all the attention you're giving this project!

renekliment:
@maso27 
After a quick look at the code, I see that the _play_audio_ is called in threads and there's no queue implemented. I think that the reason it seemed serial and working so far was that only one short Alexa message was played and loading the music streams takes a while, so it had time to get played. Now, when there's two Alexa messages, the second doesn't make it in time before the stream starts playing. This will however be fixed in that other PR.

The code is still the same - check out the [rasperrypi device file](
https://github.com/renekliment/AlexaPiNG/blob/46e9ce6142e998a3825454bfe16d8e657bb5e452/src/alexapi/device_platforms/raspberrypi.py) and in `main.py` we do
```
while platform.should_record() == True or ((thresholdSilenceMet == False) and ((time.time() - start) < MAX_RECORDING_LENGTH)):
```
Are you sure it behaves like that? Last time I checked it worked as it should.

The _10s shutdown_ was actually commented out until I've decided what to do with it. I brought it back as discussed - you can set the _command_ / _duration_ / _audio_file_ in the configuration if you want to. The _audio_file_ playback however isn't enabled yet - I'll do it later if it's not a blocker for you - I need to thoroughly think about how to do callbacks design-wise. You can use it to shutdown the pi under _alexapi_ as well. Just give _alexapi_ the sudo permission without password for your specific command (only!) and set the command in `configuration.yaml` as `sudo command`.

Thanks! I (and probably anyone else) always wanted to have a personal assistant and you guys (you, Sam, Flooie and many others) made it possible the way I want it. I'm excited to work on this. I think there's a lot of potential here and if we unify the repos and improve the code quality, instead of a lot of people doing small, often incompatible changes on their own playground, we'll gonna have one, even more awesome project! I can already see it's bringing the developers together :)

renekliment:
@alexa-pi/coreteam I'd like to merge this without rewriting it to have platforms as state machines and do that later in a separate PR. (if there are no other issues of course) 

The reason for this is that this PR provides the abstraction that enables AlexaPi to run on other machines such as our laptops (without the need to have [fakeRPiGPIO](https://github.com/ArmlessJohn404/fakeRPiGPIO) installed). Having this in _master_ will enable developers to add new platforms (as some have started some work on that already) correctly, instead of just replacing RPi stuff with other stuff. I hope this is one of the reasons for other developers (also from other forks) to participate. Also, it will take stuff from `main.py` and it will be easier to work with. It doesn't go backwards in any way and it does go forward, so there's that.

Another reason is that it's gonna take me some time to rewrite it and it's gonna take some time for review / changes and I'm not sure that this would happen in another 20 days, having this open for 20 days already. (Not judging, I'm personally gonna have not that much time as before.)

Basically these indications (not states) are there now already, this just makes it better and it doesn't matter that we're planning to rewrite this. Why not make it better in the meantime :) Actually, it makes sense to do it separately as this would do abstraction and the other PR would change logic.

maso27:
So I've been seeing long-term stability issues with this one.
I'm not entirely sure that it's related to AlexaPi, because I also started using the September Raspbian Lite image as a jumping off point, and I've had some other stability stuff happening too.

But I haven't yet had AlexaPi respond after running overnight.
The process doesn't die, and I can still ssh into the Pi (I'm using a Pi2 for this), but button press and trigger word don't work any more.

I'll try a reinstall--maybe with last May's Raspbian image--and see if it helps.

illperipherals:
Hmmmn... I am running the Pixel Raspbian update on a Pi3, and I have been pretty solid with my uptime on that device. Are you still using the PS3 Eye array for your audio? I did have some recent problems with that when I upgraded, and I reverted back to the C-Media USB for mic.

renekliment:
@maso27 
Well, it _has to_ work under the latest Raspbian, we have to make it work. What other issues did you have with the OS?

What about logs? `journalctl -u AlexaPi.service`
So you had no issues with older AlexaPi revisions?
Are you trying latest changes from this PR?

Thanks for testing!

maso27:
@illperipherals 
I'm not running the PS Eye for this.  That's on a Pi3 I have in my kitchen.  It's high-use, so I don't mess with that setup.  This is a webcam and some PC speakers I had hacked together previously.

@renekliment 
I think I'm one commit behind on this, so I'll integrate the new changes and make a new image and see what's different.
I had issues with the whole box ceasing to respond to pings a few times.  This was with CAT5, not wifi, and I had to cycle power to get it back.

I'll play more, but mostly wanted to see if anyone else was having issues with any of this stuff.

renekliment:
Alright. Three commits back I fixed Alexa (not) saying "yes" that I've broken before. Also make sure you have updated config.

maso27:
Fresh image, reinstall.  Now the button does nothing.  Is there something I need to set in a config or something?

renekliment:
@maso27 Sorry! Fixed. Just `git pull`.

maso27:
No worries.  I pulled it in, and will try it tonight when I get home.

maso27:
I think my previous image was corrupted somehow.
22 hours uptime and Alexa is still working fine.

I think this is ready to be merged.

illperipherals:
Good deal! I have been running well over a day since I updated, and everything is working fine on my end. I agree on the merge.

renekliment:
Awesome! This is a historic moment for AlexaPi! Merging!
Also creating an issue to rewrite device platforms as state machines.

PR #36, opened 2016-10-02
Fix current build errors

renekliment:
This is just a little thing before we do at least the basic CI for real. It fixes the few issues so the tests don't fail by default, which is annoying.


renekliment:
Also, after this is merged, the tests have to pass for any PR to get merged.
We'll be adding some real tests later, but now we have at least the super basic _pylint_ stuff (`--errors-only`) as a sort of sanity check.

@alexa-pi/coreteam Please tell me, if this is okay.

renekliment:
@alexa-pi/coreteam Please tell me if this is ok. It's a very small change, but it lets us start using CI, even though we'll have just very basics tests for now.

maso27:
It's ok by me.

renekliment:
Awesome, merging!

PR #41, opened 2016-10-10
Audio Queue Manager

davidmroth:
Added an audio queue manager to fix Audio Player playback issue (were multiple audio stream played simultaneously). Also imported build fixes.


renekliment:
Thank you for the PR - good idea. I've added a quick Code Review there. Please have a look at my suggestions.

Thanks!

PR #48, opened 2016-10-20
Added OrangePi compatibility

GadgetReactor:
Hi! this PR is for added OrangePi compatibility.
I modified config.yaml to recognise a device tag, but perhaps the raspberrypi option can be amended in future as well to be more generic.

The OPi has no GPIO event detect function, so the button pressed detection had to be modified somewhat. I am not able to get it to listen when entering the detect_button(channel) function. Is this working properly on the RPi?

Also, getch is not used in the program but is in the import list. Can it be removed? I have problems installing it via pip on OrangePi.


renekliment:
Hi. Thanks for doing OPi support!

There is work on device platforms (#30). Could you base your changes on that PR? That is the way to go - we don't want to have this stuff in _main.py_ with `if (device == 'blah') ...`.

On RPi it works fine.

I'll soon do a cleanup to get rid of unused things, but not yet. Please delete it only locally for the time being (don't commit it).

Also, please do a pip install of the library in `.travis.yaml` so your build passes successfully.

Thank you!

GadgetReactor:
Ah thanks. Didn't realise there was that ongoing work.

GadgetReactor:
Thanks. I was able to add OPi to the (#30). But the button detection need some work on OPi - the GPIO library there is rather limited.

renekliment:
Awesome! Feel free to make a PR against that branch in my repo (until it gets merged here), so we can see what's the status of it and can discuss it if we want to.

Issue #45, opened 2016-10-17
Switch from pyalsaaudio to pyaudio


renekliment:
We actually use this for recording only, so switching the lib will have no impact on playback.

Done in https://github.com/alexa-pi/AlexaPi/pull/180 (only in `dev` branch actually)

ImEmJay:
+renekliment Getting rid of pyalsaaudio is a good idea. if no one else has already started i would have a look at it. but do we need to support pyaudio and sounddevice? As far as i see they both wrap the PortAudio library. Snowboy and pocketspinx can both work with them.

renekliment:
@ImEmJay go ahead :-)

Note, that the code for pocketsphinx has been moved to triggers and the code around normal recording has been changed a lot - all in a PR #119. So it may be best to wait after the PR is merged, which should be today.

Probably not, I was just thinking out loud.

Whether to do pyaudio or sounddevice ...
- snowboy uses pyaudio for examples
- this speaks in favor of sounddevice: https://www.reddit.com/r/Python/comments/3k11g5/whats_a_good_sound_recording_library/cuu6gvy/

I would go with sounddevice, but it's up to you.

ImEmJay:
@renekliment I also think that pyaudio is more legacy and sounddevice is more under active development. but could have been that you know of some huge differences in platform compatibility or such ^^ but as they both just wrap i think that shouldn't really be an issue.

renekliment:
Okay. Let's go with sounddevice then.

renekliment:
For the record, the messy PRs (#119 & #110) have been merged, so there should generally be no major conflicts when editing the recording code.

renekliment:
Hey, just checking in. Any progress on this?

ImEmJay:
@renekliment hey, had no time over the last week. Think I'll get it done this weekend. But after my initial tests I would maybe change my decision and go with pyaudio. Sounddevice brings in a good amount of dependencies and propably unnecessary complexity. But will see when I finally get into it.

ImEmJay:
@renekliment Because of your comment on #152. Checking and listing devices can be done in both and was planned to add another PR to improve setting/checking the input device.

renekliment:
> unnecessary complexity

What do you mean by that exactly?

> good amount of dependencies

Do you mean NumPy? I thought it's optional and we don't have to install it.

renekliment:
[...]

Issue #37, opened 2016-10-02
[device_platforms] (new) sysfs GPIO


renekliment:
This can be done once #30 gets merged.
Might be a good idea to wait for #24 to be done.
A sketch however can be done now.

Make a new device platform that uses https://github.com/derekstavis/python-sysfs-gpio.
It will be similar to the _raspberrypi_ platform. I suppose LEDs are gonna be easy, the triggering (button or whatever) might a teeny-tiny bit tricky (debouncing, not missing it while polling the thing).

We also need to solve permissions to `/sys/class/gpio` stuff. We're gonna do it the way RPi does it. We need
- _gpio_ group
- user _alexapi_ in that group
- group of the corresponding stuff changed to _gpio_ - google _udev GPIO_; on OpenWRT there's an udev alternative

We could then use this for anything that has GPIO exposed via standard kernel interface.

I'm not sure what the advantage of _RPi.GPIO_ library and the clones is. I know it touches `/dev/mem` or `/dev/gpiomem`, but what for? Does it go around `/sys/class/gpio`?
We could even use this as an alternative for boards that have the python library (Raspberry Pi, Orange Pi, Turris router).

renekliment:
This is free to take up.

PR #55, opened 2016-10-22
Add Orange Pi support

GadgetReactor:


renekliment:
Great! For the reference, I'd like to point out, that CR has been done in renekliment/AlexaPiNG#2 and the code seems fine and is ready to get merged.

**Requests / Questions:**

1. Please add the command 
    ````
echo 'Y\n' | pip install git+https://github.com/duxingkei33/orangepi_PC_gpio_pyH3.git -v
````
in `.travis.yaml` after `pip install fakeRPiGPIO`. The tests should then pass.

2. (from that other PR) On the library page it says:
> python control orangepi_PC ext GPIO ALLwinner H3 base on pyA20 0.2.1
> open source modified base on pyA20

  Do you know if it's still compatible with the A20? Or the way to go there is just installing the original library for A20 and the same code works?

**Notes / Observations:**

1. Installation of the GPIO library needs to be done manually for now. It goes like this:
    
    ```
pip install git+https://github.com/duxingkei33/orangepi_PC_gpio_pyH3.git
```
When we do refactoring of the install script, we'll add it there. Poking around there now would just be adding mess to the mess.
2. Unfortunately this needs to be run as the user `root`, because the library works via `/dev/mem`. To do that, users must change the user in systemd service file / init file.
3. Other ways to support Orange Pi would be:
    1. via kernel's sysfs GPIO interface - `/sys/class/gpio` as in #37; if we set it right (udev rules and stuff), we could run AlexaPi as the _alexapi_ user; not sure if the legacy / mainline kernels on OPi have the neccessary module - we'd have to look
    2. via some sort of GPIO proxy, that itself would run as root so AlexaPi could run as user _alexapi_ and would communicate with the hardware via that proxy; special care would be needed here to avoid any possible bugs in the proxy

**Todo for me:**

1. Test this on my Orange Pi 2E.
2. Write the installation instructions on the wiki here (create the wiki first). Also, mention other ways to run it on an OPi than this one.

GadgetReactor:
### Response to Requests | Questions
1. .travis.yaml updated
2. I am not sure on the compatibility with the A20. I only have an Orange Pi PC Plus and Orange Pi PC. Both works with this library. But it would make sense that  just installing the original library for A20 and the same code should work, as it is more of a wrapper.

renekliment:
Ah, silly me.

1. Try this instead: 
    `yes | pip install git+https://github.com/duxingkei33/orangepi_PC_gpio_pyH3.git -v`
    We need to proceed with the

    > No processor detected
> Warning! Detected and target processor mismatch.
> Do you want to continue [Y/n]?

    and this should work better.

    Also replace `pylint --errors-only --extension-pkg-whitelist=alsaaudio src/alexapi` 
    with `pylint --errors-only --extension-pkg-whitelist=alsaaudio,pyA20 src/alexapi`
    I completely forgot about that.

2. The _fake_ library isn't needed, you don't need to bother. But If you want to go for it, make sure it works locally first. Travis CI gives
    > E:  5, 0: No name 'gpio' in module 'pyA20' (no-name-in-module)
    > E:  5, 0: Unable to import 'pyA20.gpio' (import-error)

    And I've verified locally that the import doesn't work. Also, please use the stuff from _fakeRPiGPIO_ for outputting things.

Thank you!

renekliment:
So I did a pretty extensive testing of this on my Orange Pi+ 2E (H3). It works great! I just wish I didn't waste an hour on pin numbers :disappointed: 

Everything should be [documented thoroughly](https://github.com/alexa-pi/AlexaPi/wiki/Devices#orange-pi--other-a20--h3-based-boards). If you think of anything else, feel free to add it there.

So after the build is fixed, I'll merge this.

Thanks!

P.S.: I'm getting audio distortions on the playback start / end. Do you have that too on your Orange Pis? I guess it's OS (config, drivers) or AlexaPi (buffers) related, but I'm not sure which one.

renekliment:
@ajplante Letting you know that now we have support for Orange Pis (after we merge this, which is now a only matter of setting up the build properly according to what I've written ^).

maso27:
@renekliment 
As far as audio distortions are concerned, I wonder if it's at all similar to issues I'm having with the CHIP?

On the CHIP, the first portion of all audio is clipped off ("yes" is basically a very short "s" sound) and it looks like [a number of people](https://bbs.nextthing.co/t/strange-audio-output-behavior-missing-first-few-words/5495) are seeing the same problem when working with different audio packages.

The current theory there is that CHIP's drivers put the audio chip to sleep, then wake it up when it's time to play something.  Sam got around it in his original AlexaCHIP by playing 1 second of silence before every response.

May not be related for the Orange Pi, but I thought I'd put it out there.

renekliment:
Well, the "yes" is always fine so I don't think that's it. It's mainly (and probably only) the endings. It sounds like it plays what it's supposed to and then for a split second plays some junk from some audio buffer (it's always a tiny part of some of the played messages).

I'm gonna investigate after #45. I'd hate to do something and then figure out later I have to throw it away.

maso27:
@renekliment 
I think I know what you're talking about.  You can hear it right at 0:42 on [this video](https://www.youtube.com/watch?v=nm0KH_Odq7c).  I heard it on a demo board when I was first trying to get Alexa into Arch Linux, but I haven't heard it on a Raspberry Pi, either in Arch or Raspbian.
I think it's probably hardware/driver specific.

Shoot!  I was hoping I could get you guys to debug the CHIP issue :)

ajplante:
@renekliment  
that is so awesome, are the audio distortion issues fixed?

ajplante:
I had the audio distortion issue when I managed to get the older build running on my oRangePi lite with the fakegpio module.

GadgetReactor:
Yep. I get the same audio distortion issue too: happens at the end of the audio file. The start of the audio file is fine.
I think it is to do with VLC.instance and alsa audio. When I switched to using Pulseaudio on another Snowboy build, I do not get the audio distortion.

GadgetReactor:
@renekliment could you advice on how to modify the setup for the [fakepyA20GPIO](https://github.com/GadgetReactor/fakepyA20GPIO/) ? 

It seems to work with 
``` 
git clone https://github.com/gadgetreactor/fakepyA20GPIO.git
cd fakepyA20GPIO
sudo python setup.py install
```
as compared to 
```sudo pip install git+https://github.com/gadgetreactor/fakepyA20GPIO.git" 

thank you.

renekliment:
Thanks for the update, merging!

Hopefully, we'll solve the audio issue with #45 or after we do that.

As to the library install, the second option doesn't work for me either, so I think they're equal. I don't have any personal experience with it so I can't help you with that I'm afraid. This is what I found on Google and it kinda makes sense: [[1]](http://stackoverflow.com/questions/3781522/why-do-python-modules-sometimes-not-import-their-sub-modules) and [[2]](http://stackoverflow.com/questions/8899198/module-has-no-attribute), even though we use submodules here as well and it just works. Well, I don't know.

GadgetReactor:
Thanks for the merge!

FYI, I fixed the library install issue. It's required to specify the subfolders in setup.py.

renekliment:
Btw, I believe you can delete the branch `orangepi-compat` now.

Ah, I see, wouldn't have thought to do that :-)

renekliment:
@GadgetReactor @ajplante In case you are interested I've just installed a beta image of Armbian (the Ubuntu based one) with kernel 4.9 and the audio issues are gone. So mainline is the way to go! :-) However I don't recommend it now, because proper thermal stuff isn't there, so if you don't have proper cooling and / or underclock, you could damage your board. So this is a notice that it will work in the future!

Also, it works fine under legacy kernel with SoX instead of VLC, but someone has to yet write the playback_handler.

@maso27 What kernel are you running on CHIP?

GadgetReactor:
Thanks for sharing. Will stick to Armbian first, it is really quite stable!
But will explore using other media players instead.

PR #59, opened 2016-10-22
Documentation update & Change Log (fixes #57)

### LINKED ISSUE #57: Change Log
This is actually a continuous thing, but we need to start.
I think doing it according to this: http://keepachangelog.com/en/0.3.0/ is a good idea.

The section _Project history_ from `README.md` should be moved there (rewrite it according to that format) because it belongs there and we don't want to have a mile-long README. Also, we should add the stuff that's been changed so far.

The main reason for this is that people are actually starting to use this AlexaPi repo and it's a good idea to write at least breaking changes somewhere.

renekliment:
This:
- updates tiny bits of README,
- starts a Change Log,
- deletes stuff that has been moved to the Wiki or the Change Log.

Fixes #57.


renekliment:
@alexa-pi/coreteam So this the small update. Let me know what you think.

renekliment:
Now I'm thinking we should probably move the Installation info to wiki as well?
The README would only contain the info what the project is and probably Requirements so people know what they need or should have to get it running.

renekliment:
Note to myself: Change the docs url in systemd unit file.

sammachin:
Install info in the Wiki seems like a good idea, so long as we add a link to it from the README, people won't think to look there, they'll just try it and raise an issue when something breaks!

illperipherals:
This looks good to go to me.

renekliment:
Great! Merging!

PR #64, opened 2016-10-24
Common config for device platforms

renekliment:
So it turns out we're gonna have a mile long config, if we keep adding platforms / features. This should help with that.

I should probably also: 
- [x] add a [table in docs](https://github.com/alexa-pi/AlexaPi/wiki/Device-platforms) listing what platform supports what features.

@alexa-pi/coreteam Another of the small PRs. Please have a look.


maso27:
Alrighty!  I tried this with "raspberrypi" and "desktop" with no issues.  I also had no trouble with "orangepi" but I'm using a Raspberry Pi so I don't know how valid that is.

Considering there were no errors, though, I'd say it's ready to merge.  Unless there's a reason to test this specifically on an Orange Pi.

renekliment:
Great! Merging.

PR #66, opened 2016-10-29
Configurable audio output, fixes #22

### LINKED ISSUE #22: Audio output: DOCS and CONFIG
Hi,
Solved my own issue as i was writing it up

I had no audiooutput, but i could tell from the LED that it was recording and playing back. 
Green and Red (I miss the Yellow for "transmitting)
It also would give a Green LED if i just said "Alexa"

The issue was that i had a HDMI monitor connected to set up the WI-FI. which i found out, as i'm sure  most of you guys know already, that that kills the Audio Jack output.

Might be worth a mention in the instructions for others that are like me and just starting out using Pi.

renekliment:
Very small change again.
You have to change your config file - let's break stuff every day!

### DIFF
--- a/src/config.template.yaml
+++ b/src/config.template.yaml
@@ -1,9 +1,18 @@
 # Rename this file to config.yaml and fill in the fields
 
+# See https://github.com/alexa-pi/AlexaPi/wiki/Audio-setup-&-debugging
+# to get a better idea of how to set things in here
 sound:
-  # Name of your microphone/soundcard in arecord -L
+  # Name of your ALSA microphone/soundcard in arecord -L
   # you can also try setting it to 'default'
-  device: "plughw:1"
+  input_device: "plughw:1"
+
+  # you can go here with alsa, pulse or others
+  # keep blank for the best audio output selected by the underlying code
+  output: "alsa"
+  # output device for the audio output specified above,
+  # currently supports ALSA only
+  output_device: "default"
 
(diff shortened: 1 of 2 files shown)


maso27:
This is something I was going to suggest after changing sound cards on the CHIP yesterday.

I built it and ran it, and it works with the default 3.5mm jack.

I haven't tried with any other cards yet, but have a question.  How does this do different devices on the same card?
The raspberry pi has two devices under the card name ALSA.  Is the config.yaml set up for me to enter "CARD=ALSA,DEV=0"?

renekliment:
To be completely honest here, I don't know. All is does is that if you set `sound/output_device`, it passes that to `--alsa-audio-device=`, so anything that you would pass to `aplay` for example will work here.

Well, that's sure weird. What's your `aplay -l` and `aplay -L`? On other computers I've been using `plughw:0` and `plughw:1`, but I gotta say, the ALSA device naming seems like magic to me so I always circle a few options until things start working.

renekliment:
@ShawnHarte Thanks, merging then!

Also, can you please someone test if you can set audio output to go through the analog, with HDMI connected on a Pi? I don't have any HDMI peripheral to test.

ShawnHarte:
Yes audio can run through the analog out while HDMI is attached.  plughw:1 vs 0 works.  I have PA enabled on a pi3

PR #69, opened 2016-10-30
Add C.H.I.P. platform

maso27:
This is a start for the C.H.I.P. computer.

There are still issues with the C.H.I.P. which can be solved with a little extra work:
1. CHIP_IO.GPIO needs to be built following the instructions included within chip.py
2. It doesn't like to be run as user "alexapi".  Need to change user and group to "root" within AlexaPi.service, and change "ProtectHome" from "true" to "false".
3. The first second of audio is cut off with the on-board audio, and music is scrambled.  A USB audio device is recommended.


renekliment:
Awesome! I'll have a closer look during the week, but it looks fine so far.

Few points here:
- It would be cool if you could look at #64 so we can have shorter config here.
- Please move the instructions to the documentation wiki. Follow the [Orange Pi section](https://github.com/alexa-pi/AlexaPi/wiki/Devices#orange-pi--other-a20--h3-based-boards) as an example for both library install and systemd unit override - maybe move the systemd override to a special section and link from both Orange Pi and CHIP.
- Write the note about using external audio in the wiki also.

I believe that we can aqueeze
> git clone git://github.com/xtacocorex/CHIP_IO.git
> cd CHIP_IO
> sudo python setup.py install
> cd ..
> sudo rm -rf CHIP_IO

into just `pip2 install git+https://github.com/xtacocorex/CHIP_IO.git`.
General note here - when cloning from GitHub, don't use the git protocol as it's unencrypted AFAIK - correct me if I'm wrong.

Thanks!

maso27:
Cool.  I'll look into your notes.

The install instructions posted in the notes of chip.py were copied explicitly from the other Git page, and I agree that they could be improved.  Considering they might change on that page anyway, I think adding our own improvements would be fine.

renekliment:
Great!

Yeah I know. I mean everyone uses that command sequence, but we're gonna be smart about it. It does the same thing, but it's much shorter and faster to type :)

renekliment:
Alright. Now you can use the #64 stuff to shorten your config. Check out the other platforms.

src/alexapi/device_platforms/chip.py
```diff
@@ -0,0 +1,118 @@
...
+import alexapi.bcolors as bcolors
+
+
+class ChipPlatform(BasePlatform):
+
+	def __init__(self, config):
+		super(ChipPlatform, self).__init__(config)
+
+		self.__config = config
+		self.__pconfig = config['platforms']['chip']
```
renekliment: Don't forget that if you want to use the _common_ platform config, you have to adjust this a bit - see other device_platform files.

maso27: Thanks.  Yeah, I still had some stuff to do.

I think I've captured it all though--added notes to the wiki too.  See what you think now.

renekliment:
So, good on the docs and the common config!

I'm gonna need to do a bit of a cleanup regarding device_platforms due to my bad design, but since it will touch all of them, I'll do it in a special PR.

Merging, thanks!

PR #68, opened 2016-10-30
Playback handlers + Queue

renekliment:
~~This is based on #22, therefore #22 needs to be tested and merged first!~~ DONE
Fixes #25 and #51.
Replaces #42.

So guys, this should fix _a few things_ regarding audio. This is as huge as device_platforms. This:
- will allow you to use separate audio outputs for Alexa's speech and media
- will allow you to use a different engine than _vlc_, which for some reason doesn't work for some people; so people who have had success with _mplayer_, _mpg123_ or whatever can write such handlers and use it with our upstream code

@alexa-pi/coreteam Please test (#22 first though).

By the way, after all the currently open PRs (+ @maso27's CHIP _device_platform_ file) get merged and no other issues come up, I think it's time for **version 1.3**! Let's work to get this done.

PR #89, opened 2016-11-11
device_platforms cleanup, fixes #73

### LINKED ISSUE #73: device_platforms cleanup
- Get rid of the fallback to the Desktop device_platform so users can see the errors on import and can decide to manually set another platform. The current state can be very confusing!
- Introduce something like `.indicate_error()` and `.indicate_warning()`, maybe with a _persistent flag_ so we don't have crap code like this in main.py:

	```
	platform.indicate_playback(False)
	platform.indicate_recording(False)
	for x in range(0, 3):
		time.sleep(.2)
		platform.indicate_recording()
		time.sleep(.2)
		platform.indicate_recording(False)
	```

    Also, this is crap as well:

	```
	def indicate_setup_failure(self):
		while True:
	```

    ... the platform shouldn't block the main program like this. If we need to block main, we need to do that over there.
- Unify code for device_platforms where it makes sense and unify code for device_platforms that use RPi.GPIO-like libraries, because it turns out in some cases we can use identical code.
- do `.indicate_processing()`

renekliment:
* disabled fallback to Desktop
* do indicate_success() and indicate_failure() instead of crap code
* common code for platforms and RPi-like platforms
* bring back indicate_processing()

@alexa-pi/coreteam Please test. I've tested this on my desktop only.


renekliment:
Tested this on my Raspberry Pi and Orange Pi and seems to work fine now.
@maso27 @illperipherals @sammachin Can you maybe test also on a CHIP (or other if you have time)? It should work the same, since it uses the very same code now, but just to be sure. Thanks!

maso27:
I'm running this on the CHIP with no problems.  I don't have LEDs on it, though.

Let me know if you need me to check that out too.

renekliment:
Do you have a button there? Does it work? I don't suspect there should be any issues.

maso27:
Yep. I have a button attached, and it works fine.
I had to follow all the CHIP-specific instructions on the wiki, but that's why they're there. They all worked, from a fresh installation of Linux.

renekliment:
Great! The install should be easier after #82 gets merged.

Let me guys know when you think this is ready to get merged.
@lechat This addresses some of the things mentioned in the cleanup issue. Please let me know what you think. Cheers!

lechat:
Yes, setting default value for unused parameter is better. Thanks!

renekliment:
Alright. Since it's hard to get an approval for merging, I'll ask another question: Are there any objections? :smile:

maso27:
Sorry!  No objections here.

renekliment:
Alright, I'm gonna merge this then :smile: Thanks!

This should leave only #82 pending for approval. After that's done, I'll finally finish #68 so we can release _v1.3_. 

@lechat How about your _logging_ branch? Is that ready? Since #68 is gonna take me some time, we could fit it in the release if it is ready.

PR #99, opened 2016-11-30
Hotfix for #98

renekliment:
It adds space before key name to prevent matching other words, so device doesn't match lines beginning with `output_device`, etc. This happens only on Raspberry Pi, because only there the config variable `device` gets set.

It's kind of a silly fix, but it seems to work fine.

@alexa-pi/coreteam Please have a look at this ASAP. Thank you.


renekliment:
@alexa-pi/coreteam I'm giving a 1 week deadline for objections / comments starting now and then I'm gonna merge it, because it's a one-line hotfix for an issue that can confuse new users a lot.

maso27:
Finally got to this.  Yep.  It works on the Raspberry Pi.

Thanks!

renekliment:
Great!

PR #104, opened 2016-12-10
stop the player on app exit, fixes #56

### LINKED ISSUE #56: Alexa gets messed up after Ctrl-C during playback
When playing back Tunein radio channels, I am not able to issue the Ctrl-C command. It also breaks voice detection thereafter.  Requires a Ctrl-Z and a kill -9. 

```
Sending Playback Progress Report Request...
Playback Progress Report was Successful!
```

renekliment:
Just a quick one-liner.

@alexa-pi/coreteam Please have a look, there should be no issues here. Thank you.


maso27:
Good fix!
I can break it during "News Flash" without the fix, and can't with it.

Ready to merge.

renekliment:
> I can break it during "News Flash" without the fix, and can't with it.

Just to be sure, by _break it_ you mean _break_ the app, right? So this fixes it.

maso27:
Haha yeah sorry.
A little ambiguous there.
I could break AlexaPi before by typing "Ctrl-C" when the News Briefing was playing.  Got stuck, wouldn't exit.

I changed this line, and now "Ctrl-C" exits the app cleanly instead.

So this fixes it.

renekliment:
Great, thanks :-) Merging then.

PR #110, opened 2016-12-15
Use python logging instead of prints

lechat:
Here is an examples on how logs look like (note that I don't have any audio hw connected to my RPI now, so all logs will show failures)

Command line with `--debug` option:
command_line

/var/log/syslog:
syslog

journalctl:
journalctl

I've added command line parameter `--daemon` and updated service and init.d script to use it. This is required to disable timestamps log messages when code runs as daemon.


renekliment:
Great! Looking forward to merging this!
I'm giving myself a break from being active in AlexaPi until the _v1.3_ release (bugfixes only for now), so I'm gonna test this and merge first thing after that.

Two little things from a brief look:
1. Could you fix `tunein.py` so it uses the common logger and does not log into `tunein.log` which creates a mess?
2. Please load the default log level from config (override-able from CLI) so people can set their default desired level and can set it for AlexaPi running as a daemon.

Thanks again, good job :smile:

renekliment:
So I've actually tested this. It works fine. I've got a couple notes / TODOs:
- [ ] Please put a note in the Configuration changes on the wiki about installing `coloredlogs` (or re-running setup)
- [ ] In `silence_listener`, please remove the phrase `Debug: ` in the log messages (3 of them), since it's defined by the method called. It's a remnant from the _prints_.
- Man, the colors are crazy :smile: Let's keep it like this. If people have issues, we can easily turn it off.

That's it from me and I'm ready to merge this after the review comments are addressed.

src/alexapi/device_platforms/dummy.py
```diff
@@ -11,36 +14,28 @@ def __init__(self, config):
 		self.should_confirm_trigger = False
 
 	def setup(self):
-		if self._pconfig['verbose']:
```
renekliment: Since you got rid of the platform config option _verbose_, delete this:

```
    # whether to print messages when methods are called
    verbose: false
```

from the config template. Be sure to keep the `dummy:`, so the config loads successfully.

renekliment:
Also, please do the PR against the `dev` branch. Thank you.

src/alexapi/device_platforms/dummy.py
lechat: I think that it makes sense to make "logging" option in config file, so anybody can set it permanently to DEBUG, if they want to. Default will be INFO and `-d` option for `main.py` will override it at runtime.

lechat: The reason I removed it from dummy is that no other platform module had such option, but they should. It is actually better to move that logging to `baseplatform`, so it can be enabled for any kind of platform.

renekliment:
I've changed the branch and I'm reopening this for the practice. I hope you don't mind.

lechat:
Sure. I'll work on remaining changes this coming week, once all holidays are over. ;)

lechat:
Fixed all points and rebased on `dev` branch.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ This is mainly a test of doing bugfix releases.
     - Default volume
 - Different audio output devices for speech and media can be specified.
 - Support for Arch Linux.
+- Added dependency on _coloredlogs_
```
renekliment: This should be all under [Unreleased].

CHANGELOG.md
```diff
@@ -49,7 +50,8 @@ This is mainly a test of doing bugfix releases.
 - There is no default command for the _long_press_ feature.
 - Abstracted device platform code into **_device_platforms_** which means we can now support other devices within the same codebase and users can now write their own independent device platform files.
 - Abstracted playback library into **_playback_handlers_** which means we can now support multiple libraries within the same codebase and users can now write their own independent handlers and can route their sound through whatever they want to.
-
+- Changed logging to use Python logging
+- Added logging configuration to convig.yaml
```
renekliment: typo convig.yaml

src/main.py
```diff
@@ -42,13 +60,29 @@
...
+	log_level = logging.DEBUG
+else:
+	log_level = logging.INFO
+
+if cmdopts.daemon:
+	coloredlogs.DEFAULT_LOG_FORMAT = '%(levelname)s: %(message)s'
+else:
+	coloredlogs.DEFAULT_LOG_FORMAT = '%(asctime)s %(levelname)s: %(message)s'
+
+coloredlogs.install(level=log_level)
```
renekliment: Where do you extract the setting from the config?

renekliment:
Also, I believe we can now delete the `bcolors.py`?

CHANGELOG.md
```diff
@@ -49,7 +50,8 @@ This is mainly a test of doing bugfix releases.
 - There is no default command for the _long_press_ feature.
 - Abstracted device platform code into **_device_platforms_** which means we can now support other devices within the same codebase and users can now write their own independent device platform files.
 - Abstracted playback library into **_playback_handlers_** which means we can now support multiple libraries within the same codebase and users can now write their own independent handlers and can route their sound through whatever they want to.
-
+- Changed logging to use Python logging
```
renekliment: This should be under `[Unreleased]` too under an appropriate section.

src/main.py
```diff
@@ -73,7 +73,7 @@
 if debug:
 	log_level = logging.DEBUG
 else:
-	log_level = logging.INFO
+	log_level = logging.getLevelName(config.get('logging'.upper(), 'INFO'))
```
renekliment: `.get` for default value - sweet! We might use that somewhere else also. I'll think about it.

I do believe however that `.upper()` should be in a different place.

renekliment:
And join our Gitter chat room :-)

src/main.py
lechat: Ah, yes! Stupid mistake. :)

renekliment:
Okay, I do have one more thing :sad:

When I run with no config files I get:
> `No handlers could be found for logger "alexapi.config"`

which is due to the fact that the logging setup happens after the `alexapi.config` module is imported and that check happens on module import. Changing `logger.` to `logging.` works. 

But then I made a project-wise search and there's plenty of both `logger.` and `logging.` even in `main.py`. Can we make this consistent somehow where possible? I guess only the `__init__.py`s have to use `logging.`? Maybe moving this into a module (like `alexapi.logging`) and importing before `alexapi.config` would work so we only have `logger.`? And set the log level from config after we have config? I'm not sure what's the best approach - just giving ideas.

I truly believe this is the very last thing and then we can finally merge this.

Thanks!

src/alexapi/device_platforms/dummy.py
ImEmJay: I also use the _verbose_ option in the unmerged `hyperion` platform. Shall we keep it in the specific platform config or move it to a more general place?

renekliment: @ImEmJay In your platform, use `logger.debug()` (or `logging.debug()`, not sure).

ImEmJay: @renekliment I do actually use normal/debug output (and will use the `logging.debug()` after rebase/rework). I just used the _verbose_ option to output additional info (the JSON request to hyperion). If we got completely rid of that option I can also include that output in the normal debug output.

renekliment:
Rebasing this on `dev`.

src/alexapi/triggers/pocketsphinxtrigger.py
```diff
@@ -32,9 +35,8 @@ def setup(self):
 		ps_config.set_string('-keyphrase', self._tconfig['phrase'])
 		ps_config.set_float('-kws_threshold', float(self._tconfig['threshold']))
 
-		# Hide the VERY verbose logging information
-		# TODO: temporary here until the logging PR is merged - check if we need to change this
-		if not self._config['debug']:
+		# Hide the VERY verbose logging information when not in debug
+		if logger.getEffectiveLevel() != logging.DEBUG:
```
renekliment: This doesn't actually work and I wasn't able to find a working solution. Please fix. Thank you.

renekliment:
Alright. Rebased on `dev`.

Please have a look at my previous comment and fix finding out the log level, because I didn't know how. Then we'll merge this and the two other PRs can follow.

src/alexapi/triggers/pocketsphinxtrigger.py
```diff
@@ -36,7 +36,7 @@ def setup(self):
 		ps_config.set_float('-kws_threshold', float(self._tconfig['threshold']))
 
 		# Hide the VERY verbose logging information when not in debug
-		if logger.getEffectiveLevel() != logging.DEBUG:
+		if logging.getLogger().getEffectiveLevel() != logging.DEBUG:
```
renekliment: Hm, this still doesn't work for me :-(

src/main.py
```diff
@@ -87,20 +80,35 @@
 silent = cmdopts.silent
 debug = cmdopts.debug
 
+config_exists = alexapi.config.filename is not None
```
renekliment: I don't think we need this. If there is no config, it fails in alexapi.config in `__init__.py`.

renekliment:
Thank you for the changes. I see you did a merge instead of force pull. It's good we'll squash the PR when merging :smile: Anyways, still a couple of notes and the `logging.` / `logger.` stuff in other files. Thank you for looking at this.

renekliment:
Dope, confirming that setting the debug level now affects pocketsphinx.

Ready to merge this when the `logging.` / `logger.` inconsistencies are resolved.

src/main.py
```diff
@@ -528,7 +572,7 @@ def trigger_process(trigger):
 				if os.path.isfile(file_path):
 					os.remove(file_path)
 			except Exception as exp: # pylint: disable=broad-except
-				print(exp)
+				logging.warning(exp)
```
renekliment: `logger.` here?

renekliment:
@lechat Please update the config changes section on wiki.

Issue #107, opened 2016-12-12
Flash Briefing Issues


mochman:
**Your OS (including version) where you are running AlexaPi:**
```
Raspbian GNU/Linux 8 (jessie)
```

**Your hardware platform and model you are running on:**
```
raspberry pi 3
```

**Python release (`python2 --version`):**
```
Python 2.7.9
```

**Description of problem:**
```
When running AlexaPi under daemon mode (tested with both systemctl & classic).  When trying to get the "Flash Briefing" by asking "What's New", Alexa replies with "Here's your flash briefing" then goes back to idle.  When running AlexaPi from the command line with "/opt/AlexaPi/src/main.py -d", Alexa works normally by saying "Here's your flash briefing." the continuing with the brief  "Today in the news....".

When Alexa fails, this appears in the logs:

python[6233]: [718094a0] filesystem access error: cannot open file /tmp/AlexaPi-runtime-RaKTUW/DailyBriefingPrompt.750758e8-2ac8-4568-bc4a-7add1a025a1a:ChannelIntroduction:0.
python[6233]: [718094a0] core access error: File reading failed
python[6233]: [718094a0] core access error: VLC could not open the file "/tmp/AlexaPi-runtime-RaKTUW/DailyBriefingPrompt.750758e8-2ac8-4568-bc4a-7add1a025a1a:ChannelIntroduct
python[6233]: [71610f30] core input error: open of `file:///tmp/AlexaPi-runtime-RaKTUW/DailyBriefingPrompt.750758e8-2ac8-4568-bc4a-7add1a025a1a:ChannelIntroduction:0.mp3' fai
python[6233]: [71610f30] core input error: Your input can't be opened
python[6233]: [71610f30] core input error: VLC is unable to open the MRL 'file:///tmp/AlexaPi-runtime-RaKTUW/DailyBriefingPrompt.750758e8-2ac8-4568-bc4a-7add1a025a1a:ChannelI

Looking at the /tmp folder, whenever AlexaPi runs, it creates a folder "AlexaPi-runtime-XXXXXX" chmodded to 700 owned by alexapi.  

Inside that folder, after I ask for the briefing, I see the files appear then dissapear after AlexaPi can't find them.

Running "sudo --user=alexapi /opt/AlexaPi/src/main.py -d" work as it should.   

Also, asking Alexa for the weather, my calendar, general questions works normally.  This problem only seems to occur when AlexaPi has to cycle through multiple audio streams.
```

**Expected:**
[...]

PR #111, opened 2016-12-16
Fix for #107

renekliment:
We can't delete the stupid files after the processing, because _media_ files are processed asynchronously in a different thread and currently, we delete the files before they can be played.

The processing stuff needs to be refactored. We probably need to do that all in a thread with a blocking _play_ and then we can delete all the files successfully. I don't know - it's one of the zillion things that needs a rework. Non-blocking play can then be removed from playback_handlers.

@alexa-pi/coreteam @mochman Please test this - it's another small bugfix.


maso27:
So is the fact that I couldn't reproduce the problem related to a pile of speed-related issues that compounded?

Like, Pi 2 vs. Pi 3, size of the audio file for News Briefing, and speed of SD card?  Maybe with a little speed of internet connection thrown in?

I'm just surprised that it's happening for some and not others.

If this is a fix for it, great work finding it, @renekliment!

renekliment:
I'm not sure why you weren't able to reproduce it if you weren't running in the debug mode. I reproduced it simply by running `python2 src/main.py`. I don't think it's caused by a race condition or anything you mention. I don't know - a mystery.

Well, it wasn't hard to find it, because I introduced the bug :smile:

renekliment:
Reported as working by the person who opened the issue: https://github.com/alexa-pi/AlexaPi/issues/107#issuecomment-267767324

Just reverting a previously working behavior. Merging then.

PR #119, opened 2016-12-29
Triggering refactor

### LINKED ISSUE #24: Refactor: triggering

renekliment:
This fixes #24.

It is a very important PR project-wise for these two reasons:
- All the triggers can be enabled / disabled, therefore the voice activation is now optional and once this is released (in _v1.4_), we can recommend **all** the users of the original Sam's repos AlexaPi and AlexaCHIP to use the version from our repo. This is the one _feature_ that we've been missing so far and I believe there are people who don't want voice activation / the recognition engines are not available for their platform.
- Since all the triggers are now modules, we can very, **very** easily add support for **snowboy** now. Yes, we got there! We need to be smart about how we distribute it though. It's not on pypi (yet?), so we might go for a git submodule. Also, it's a C library and it's not available for all the platforms, so the task is not that trivial. Of course just dumping the Raspberry Pi tarball in the repo is a way to do it (like all the other forks of AlexaPi with snowboy), but I believe this repo is all about doing things as right as possible. 
- Which brings me to the other thing that once we have snowboy, a substantial number of the forks become obsolete and with all the added features and fixes here, we'll have most of the "merge all the forks" task covered. After that, I'd like to ask all people with the forks to delete / redirect their repos here and fork our repo instead.

TODO list:
- [x] update ChangeLog (part of this PR)
- [ ] update wiki section on config changes
- [x] change trigger passing
- [x] do _continuous_ (button press only) and _continuous-vad_ (button forces, VAD is active regardless) as described in https://github.com/alexa-pi/AlexaPi/pull/119#issuecomment-269897961


renekliment:
@alexa-pi/coreteam Please test / review.

src/config.template.yaml
```diff
@@ -48,6 +48,7 @@ triggers:
     voice_confirm: true
 
     phrase: "alexa"
+    threshold: 1e-5
```
lechat: We shall have a comment here about such values and their meanings.

dgonano:
I'm getting the following error when using the button callback trigger 
`
Exception in thread Thread-21:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "src/main.py", line 512, in trigger_process
    silence_listener(VAD_THROWAWAY_FRAMES, force_record_callback=continuous_callback)
  File "src/main.py", line 420, in silence_listener
    inp = alsaaudio.PCM(alsaaudio.PCM_CAPTURE, alsaaudio.PCM_NORMAL, config['sound']['input_device'])
ALSAAudioError: Device or resource busy [plughw:1]
`
* multiple sphinx activations work 
* button callback worked the first time ever but then fails, even after AlexaPi restart or system reboot
* voice activation via sphinx still works (after AlexaPi restart obviously as the program has crashed)

dgonano:
What is the continuous vs one-shot ?

src/alexapi/triggers/platformtrigger.py
```diff
@@ -0,0 +1,41 @@
...
+
+	type = triggers.TYPES.OTHER
+
+	_platform_continuous_callback = None
+
+	def __init__(self, config, trigger_callback):
+
+		super(PlatformTrigger, self).__init__(config, trigger_callback, 'platform')
+
+		self._event_type = triggers.EVENT_TYPES.ONESHOT if self._tconfig['event_type'] == 'oneshot' \
```
dgonano: I cannot see `_event_type` anywhere else. Should this be `event_type` to match the parent class?

src/main.py
```diff
@@ -495,75 +475,68 @@ def silence_listener(throwaway_frames):
...
-				player.stop()
+	# clean up the temp directory
+	if not debug:
+		for some_file in os.listdir(tmp_path):
+			file_path = os.path.join(tmp_path, some_file)
+			try:
+				if os.path.isfile(file_path):
+					os.remove(file_path)
+			except Exception as exp: # pylint: disable=broad-except
+				print(exp)
```
dgonano: I'd put the cleanup after everything has been done as to not slow down the response process. Maybe just before triggers.enable()?

src/main.py
```diff
@@ -495,75 +475,68 @@ def silence_listener(throwaway_frames):
...
-	global player
+trigger_thread = None
+def trigger_callback(trigger_name):
+	global trigger_thread
 
-	while True:
-		record_audio = False
+	triggers.disable()
+
+	trigger_thread = threading.Thread(target=trigger_process, args=(trigger_name,))
```
renekliment: No. It has to be a tuple, hence the comma.

src/alexapi/triggers/platformtrigger.py
```diff
@@ -0,0 +1,41 @@
...
+			else triggers.EVENT_TYPES.CONTINUOUS
+
+		self._state = False
```
renekliment: Yep, you're right.

src/main.py
dgonano: Ahk, that makes sense haha

renekliment:
@lechat @dgonano Is it okay to put only a link ( https://github.com/alexa-pi/AlexaPi/wiki/Triggers ) in the config for the options? I'm not sure I can squeeze it in less words and it seems like a lot for the config file.

Also, I remembered that I wanted to do _continuous-vad_ (like the current `master` and current _continuous_) and _continuous_ (no VAD - it records exactly as long as the button is pressed). I'll do that soon hopefully.

src/main.py
```diff
@@ -495,75 +475,68 @@ def silence_listener(throwaway_frames):
...
-		inp.setrate(16000)
-		inp.setformat(alsaaudio.PCM_FORMAT_S16_LE)
-		inp.setperiodsize(1024)
 
-		while not record_audio:
-			time.sleep(.1)
+def trigger_process(trigger_name):
+	global player
+
+	trigger = triggers.triggers[trigger_name]
```
dgonano: I know what your saying but i'd maybe do it a little different for reliability. 
in the trigger callback i'd check to see if it exists, extract it with this line, and pass the actual trigger into `trigger_process` when you start the thread. 
By this time you know everything will be fine. 
If the trigger isn't found (who knows why) you can return something meaningful etc during the callback. 

Just an idea, you don't have to.

dgonano:
@renekliment Does continuous trigger not interfere with the long press?

src/main.py
renekliment: Unfortunately that doesn't work with the current code - #107.

renekliment:
@dgonano No, it doesn't. Or to better put it: it shouldn't :smile:

src/config.template.yaml
```diff
@@ -35,8 +35,20 @@ alexa:
...
+
+    # one of: oneshot, continuous
+    event_type: "oneshot"
+
+  pocketsphinx:
+    enabled: true
+    voice_confirm: true
+
+    phrase: "alexa"
+    threshold: 1e-5
```
dgonano: just saw the triggering wiki

src/config.template.yaml
```diff
@@ -35,8 +35,20 @@ alexa:
...
 
-sphinx:
-  trigger_phrase: "alexa"
+triggers:
+  platform:
+    enabled: true
+    voice_confirm: false
+
+    # one of: oneshot, continuous
+    event_type: "oneshot"
```
dgonano: just saw the triggering wiki

dgonano:
@renekliment yeah thats fine to put a link in the config as far as i'm concerned

src/alexapi/triggers/platformtrigger.py
renekliment: Yes, it should. Thanks.

src/main.py
renekliment: Yes, that makes sense.

renekliment:
The small things are addressed and I'll change passing the trigger later.

As to the error, I can't reproduce it. I've tried on both PulseAudio and ALSA to make sure there isn't some overlap somewhere. Can you provide the whole log / dig deeper? Thanks.

src/main.py
renekliment: So I've done that and then I thought - would it make more sense to go even further so that the trigger would pass its own instance? Like
`self._trigger_callback(self)`
and 
```
def trigger_callback(trigger):
	global trigger_thread

	triggers.disable()

	trigger_thread = threading.Thread(target=trigger_process, args=(trigger,))
	trigger_thread.setDaemon(True)
	trigger_thread.start()
```

dgonano:
Yeah I'll look into the error and see if i can get it working

src/main.py
dgonano: yeah thats even better i think. no point having to retrieve it when you already have it.

renekliment: Alright. Done.

dgonano:
@renekliment So I looked into the that error a bit more. Appears to be a bit of a bug. 
It was throwing that error on plughw:1 unavailable so I tried disabling the voice trigger and everything worked fine for the platform trigger. 
I suggest perhaps the microphone is not being released properly from sphinx so that it can do the recording? Seems like the obvious answer haha

renekliment:
@dgonano Okay, I'll look into it.

renekliment:
@dgonano The bug should be fixed now. Please check.

dgonano:
Alright, I'll test today

dgonano:
just resolving merge conflicts and going to push to your branch... assuming you've set permissions to allow it.

dgonano:
@renekliment I've fixed all the conflicts and another little error that was causing an issue. Everything is working properly for me on this branch now. 
What is the command to push to the PR?
`git push upstream pull/119/head` or something?

renekliment:
@dgonano I don't know - I edited the `.md` file from GitHub before. Push to a branch in your repo and I'll pull from you.

renekliment:
Also, I'll move the _long_press_ into triggers after I pull from you. That will be the last change.

renekliment:
Alright, done - the _long_press_ stuff moved and fixed the usage of `force_record` for the Desktop platform. 

Totally, finally ready from me. If this got any better, the universe would literally explode.

@alexa-pi/team Let me know if this can be merged (CR & testing-wise).

If this gets merged before the logging stuff, I can help out with the conflict resolving.

renekliment:
@dgonano Ah, silly me. Thanks!

renekliment:
@dgonano Let me know if this is ready for the merge. I could then merge it and rebase the logging stuff so @lechat doesn't have to.

dgonano:
Yeah it's good, that why I did the approve :p 
Go for it

renekliment:
Great, thanks! Merging.

PR #116, opened 2016-12-27
Fix calling unimported function in setup (dummy platform)

renekliment:
Silly thing. Doesn't fail install, but produces an ugly warning:

> `./AlexaPi/src/scripts/setup.sh: line 191: install_device_config: command not found`

TODO list:
- [x] mention this in the CHANGELOG! (part of this PR!)

PR #123, opened 2017-01-01
Fix calling unimported function in setup (dummy platform)

renekliment:
Replaces #116 - PR against `master`.


renekliment:
@dgonano Do you approve of the PR itself then?

renekliment:
I can actually change the base - damn! Going back to the previous PR.

dgonano:
yeah i was going to say you could do that haha

PR #109, opened 2016-12-14
Improve shairport-sync integration

dgonano:
This PR allows music playing through shairport to be,
* muted,
* reduced by x%, 
* do nothing,
 
while Alexa is listening, processing and speaking. 
It reverses the set command to return to previous volume once Alexa interaction finishes. 

Main.py
* removed redundant while loop
* added start & stop interaction commands if shairport is enabled 

setup.sh
* Add section to correctly configure audio (once decision is made on to keep the config file or use a new one)
* update config.yaml with enabled = true/false
* update asound.conf so ALSA has the new soft device
* update shariport-sync.conf so it points audio to the new soft device for control 

Uninstall has been tested to. it comments out the changes made to shairport-sync so that shairport sends audio to the "default" device and is not controlled by AlexaPi

PR #125, opened 2017-01-04
PRE & POST interaction commands

dgonano:
Replacing PR #109 with more versatile implementation


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Added
+- PRE and POST interaction commands
```
renekliment: Write here something like "this enables to lower the volume of shairport-sync when talking to Alexa"

src/config.template.yaml
```diff
@@ -64,6 +64,23 @@ platforms:
...
+    interaction_commands:
+
+      # Example used to lower shairport-sync volume during interaction with alexa
+      #     pre_command: "amixer set shairportvolume 10%-""
+      #     post_command: "amixer set shairportvolume 10%+""
+      # Note this assumes you have redirected shairport-syncs audio to a seperate softdevice
+      # with control "shairportvolume". Example asound.conf avalible in the scripts folder
+
+      # Command to run after a trigger is detected but before an interaction begins
+      pre_enabled: false
```
renekliment: Lose the `_enabled` variables. If people want to disable it, they can just comment it out. If the `_command` is there and non-empty, we run it.

CHANGELOG.md
dgonano: it doesn't though, it's merely a blank command anyone can put what they want. In the config i've added shairport as an example command. 
It's disable by default because it requires setup of shairport and alsa. It's like this because you didn't want to be specific about shairport

src/config.template.yaml
dgonano: sure

dgonano:
@renekliment should I add a command to run at startup as well? 
It could be useful to make sure whatever is being controlled is in it's default state, e.g. shairport volume is 100 and now lower due to exiting during a interaction

CHANGELOG.md
renekliment: Ah, sorry. I meant "this enables for example ..." like to give people idea what they can do with it (like it's in the config template).
I didn't want to do shairport-specific code. This here should be just a comment to give an idea.

renekliment:
I like this - just a few small comments.

Yes, that makes sense. Then maybe rename the `interaction_commands` section? Something like _event_commands_? And then _pre-interaction_, _post-interaction_, _startup_ ... I don't know.

Also, please put in the wiki a short tutorial on how to set up the shariport/ALSA stuff - basically what the shell script did.

src/config.template.yaml
```diff
@@ -64,6 +64,28 @@ platforms:
...
+
+      # Example used to lower shairport-sync volume during interaction with alexa
+      #     startup_cmd: "amixer set shairportvolume 100%"
+      #     pre_interaction_cmd: "amixer set shairportvolume 10%-""
+      #     post_interaction_cmd: "amixer set shairportvolume 10%+""
+      # Note this assumes you have redirected shairport-syncs audio to a seperate softdevice
+      # with control "shairportvolume". Example asound.conf avalible in the scripts folder
+
+      # Command to run at the start of AlexaPi
+      startup_cmd: ""
```
renekliment: Please lose the `_cmd` everywhere - it's in the _something_commands_ section.

src/main.py
```diff
@@ -50,6 +51,26 @@
 if 'debug' not in config:
 	config['debug'] = debug
 
+# Get event commands
+startup_cmd = False
```
renekliment: Please do a dict here, so we have event_commmands['startup'] or similar.

dgonano: i could...

src/main.py
```diff
@@ -50,6 +51,27 @@
 if 'debug' not in config:
 	config['debug'] = debug
 
+# Setup event commands
```
renekliment: I just now realized that you can replace all this by 

```python
# Setup event commands
event_commands = {
	'startup': False,
	'pre_interaction': False,
	'post_interaction': False,
	'shutdown': False,
}

if 'event_commands' in config['platforms']['common']:
	event_commands.update(config['platforms']['common']['event_commands'])
```

renekliment: No. Empty string in python gets evaluated to `False`. (I only found out now myself)

renekliment: Or we trim the values like this:
```
config_event_commands = {k: v.strip() for k, v in config['platforms']['common']['event_commands'].items()}
event_commands.update(config_event_commands)
```

but I'm not sure if that wouldn't be over-complicating things.

dgonano:
@renekliment alright there we go, lets merge that and start closing some of these PRs :p

renekliment:
@dgonano Was about to test and I've noticed the config is under platforms/common ... Why not root?

renekliment:
And now I'm thinking the _long_press_ feature should be moved under triggers/platform. That will get rid of the platforms/common section. I'll do that in my PR.

renekliment:
@dgonano And I agree that we should merge things ASAP, because this is gonna be merge hell already (and there are probably people doing something new also). There's just small things to do everywhere.

renekliment:
@dgonano Just a couple small notes for the future:
- Let's try to honor the changes ping-pong. I know it takes a long time, but it prevents stuff like when you do a change that you think is final, but there is actually something else the tester / reviewer wants you to do. Like fixing an invalid config in this case - please see #132, so we can merge it straight away. I was also lazy and I've been doing CR only and I've tested it only now after it was merged, so it actually went in peer-untested.
- It's not critical, since it's not the `master` branch, but still. 
- When squashing, remove non-relevant messages.

dgonano:
> Let's try to honor the changes ping-pong. I know it takes a long time, but it prevents stuff like when you do a change that you think is final, but there is actually something else the tester / reviewer wants you to do. Like fixing an invalid config in this case - please see #132, so we can merge it straight away. I was also lazy and I've been doing CR only and I've tested it only now after it was merged, so it actually went in peer-untested.

Yeah noted, I would normally but i thought it was silly simple so I didn't bother. As we now know, no change is too small lol 

> It's not critical, since it's not the master branch, but still.

Another benefit of a dev branch :p 

> When squashing, remove non-relevant messages.

noted, will do

renekliment:
> As we now know, no change is too small lol

Yeah, I learned the _hard way_ too :smile: 

> Another benefit of a dev branch :p

Yep :wink:

renekliment:
Please update the wiki/config changes - add a bullet to a list to say something like "add the event_commands section".

dgonano:
Done

Issue #113, opened 2016-12-26
Adding Hyperion support for playback/record status


penfold42:
I want to use Hyperion (ambient lighting) to display the playback record status.

It's a really simple json request over tcp to set the led colours.

Should I do this as a separate platform or could we do this for all platforms ?

renekliment:
Hey. Currently, the best approach is indeed a new _device_platform_. You can inherit from whatever existing platform you like (if you want to keep its function) and just add your code.

renekliment:
@penfold42 Do you have any further questions / help requests / code to share?

ImEmJay:
@renekliment Because I also like to have Hyperion support and your 'device_platform' makes it really easy I already started implementing it.

@penfold42 I think I'm ready to push a first into a public branch in my repo within the next days. Would you like to do some testing then?

penfold42:
Yep - I have to get it working again tho which is why inhavemt done anything yet.

I was going to NOT use Hyperion-remote by rather use the json interface over tcp to set the colour.
We should expose the actual json string in the config file too

ImEmJay:
You can already set the colors independent for each state. Also you can choose between static color or 'knight rider' effect for all states. Additionaly you can override each state t use flashing notifications as you can see in the video when alexa is triggering the playback state.
I think not all effects would look good as notification and it's a bit tricky cause they all have different parameters. But im looking to integrate/customize more of the default effects. Still i would avoid to just pass the full args string as a config option.

renekliment:
This was fixed in `dev` by #127.

PR #127, opened 2017-01-04
Add Hyperion 'device_platform' for status visualization (#113)

ImEmJay:
Add support for Hyperion Ambient Lightning Software [hyperion-project](https://hyperion-project.org).

* Add new 'hyperion' 'device_platform'.
* Connect to 'hyperion' service through JSON websocket.
* Use threading for websocket communication to avoid blocking 'AlexaPi'
* Visualize 'playback', 'processing' and 'recording' state.
* Add static color visualization mode.
* Add custom 'knight Rider' effect as default visualiztion mode.
* Add optional 'Strobe' effect as flash notification.
* Support custom flash notification setting for each state.
* Support custom color settings for each state.
* Support custom speed setting for default effect.
* Support custom duration and frequency settings for flash notifications.
* Allow to disable visualization for 'playback' state.

TODO list:
- [x] Update CHANGELOG.md
- [x] Update wiki - add the new device_platform
- [x] Change triggers to reflect changes from PR #119 
- [x] Change logging to reflect changes from PR #110


.travis.yml
```diff
@@ -11,6 +11,7 @@ before_install:
 install:
   - pip install -r src/requirements.txt
   - pip install -r src/dev-requirements.txt
+  - pip install websocket-client
```
renekliment: Put this in `src/dev-requirements.txt`.

dgonano:
Haven't tested anything but it looks fine from a brief look also. Just whats been mentioned the the changelog/wiki

.travis.yml
ImEmJay: @renekliment Sorry. I thought i responded already via an app, but seems that one did not work out.

ImEmJay:
TODOs done

src/alexapi/device_platforms/hyperion.py
```diff
@@ -0,0 +1,160 @@
...
+				on_error=self.on_socket_error)
+		self.socket_thread = threading.Thread(target=self.socket.run_forever)
+		self.socket_thread.daemon = True
+		self.socket_thread.start()
+
+	def on_socket_close(self, ws): # pylint: disable=unused-argument
+		logger.debug('Closing connection')
+
+	def on_socket_error(self, ws, error): # pylint: disable=unused-argument
+		logger.debug(error)
```
renekliment: This should probably be `logger.error`.

src/alexapi/device_platforms/hyperion.py
```diff
@@ -0,0 +1,160 @@
...
+			self.socket.send(json.dumps(message))
+		else:
+			logger.warning("Unable to send Hyperion state update")
+			self.init_connection()
+
+	def indicate_failure(self):
+		pass
+
+	def indicate_playback(self, state=True):
+		if self._pconfig['indicate_playback']:
```
renekliment: Why is only playback indication configurable?

ImEmJay:
For the error/debug: there is also a connection check at startup and a
warning when alexaPi is unable to send the the request when connection is
lost. So used debug here because because even if there is a short error
(restarting your Hyperion instance or have the light turned of for some
time) it does not stop AlexaPi itself from working. But could be changed.to
an error.
For the playback config: Was just a decision by me because you either have
the sound response for that state or just pressed the trigger button. So
for me it was important that the recording processing state always have
some indication. But as said, was just a personal decision and can also be
changed to allow more customization.

René Kliment <notifications@github.com> schrieb am Di., 10. Jan. 2017,
06:38:

src/alexapi/device_platforms/hyperion.py
ImEmJay: There is also a connection check at startup and a
warning when AlexaPi is unable to send the the request when connection is
lost. So used debug here because even if there is a short error
(restarting your Hyperion instance or have the light turned of for some
time) it does not stop AlexaPi itself from working and i thought it would be annoying when you have a log full of errors when you turn your your light off over night. But could be changed to
an error.

src/alexapi/device_platforms/hyperion.py
ImEmJay: Was just a decision by me because you either have
the sound response for that state or just pressed the trigger button. So
for me it was important that the recording and processing state always have
some indication. But as said, was just a personal decision and can be
changed to allow more customization.

lechat:
Looks good to me!

ImEmJay:
Thanks for the merge

renekliment:
Merged. Thank you for the contribution.

PR #132, opened 2017-01-06
Fix config & validate on Travis CI

renekliment:
This fixes an indentation error introduced in 43941e362f664ae2d762acbdf283a5cd9ae491e3 and runs a simple check on Travis CI so this doesn't happen anymore. Along with the scripts PR, we should have the basic sanity / syntax check now.


dgonano:
Yeah I saw this soon after and fixed it in the triggering one. Oh well at least we learnt to put a check in.

renekliment:
Yep, I didn't think of it sooner. Now I hope we check _really_ everything :smile:

Issue #43, opened 2016-10-15
Get rid of memcached


renekliment:
From what I see, we use it only to store the `access_token` and it makes no sense to force the user to install it and have it running.

We can easily replace it with a local sqlite database, which we can later use to also store timers, alarms, etc. The main thing here is to carefully think about where to store the DB file.

@sammachin Is there any particular reason for why you've chosen _memcached_ originally?

sammachin:
Reasons I used memcache were:
)1 I was familiar with it as used it on lots of other projects
2) the access token expires after 1hr so when I set it I set an expiry of
just under (3590s IIRC) so it's a fairly easy way to check for an access
token and if there isn't one or it has expired it requests a new one with
the request token.
I'd rather not have to do our own time calaculations to determine if a
token has expired and making the request to find the token has expired then
generate a new one and repost the request would be a huge increase in delay.

renekliment:
I see. So it basically makes no sense to store it anywhere (permanently). There is not much reason in doing this so someone that restarts AlexaPi can keep using the existing token. It might make sense for developers, because they might restart it often, but I personally don't have _memcached_ installed on my dev machine so it probably fetches it every time, but it works without a problem.

Let's just store it in a variable together with the timestamp it was acquired. The comparison of the time is easy and we won't have to do requests with an expired token. We'll just expire it locally if the time has passed and get a new one.

sammachin:
Ok makes sense yeah store it locally with the time stamp and check that
it's under 3590s old.

The only scenario where mc helps is if you restart the python app but not
the system it will persist but like you said that's an edge case and
fetching a new access token isn't hugely expensive on its own, although I
expect AVS wouldn't like it if we all did it on every request. Limiting the
amount the refresh token is passed over the wire also helps with security

renekliment:
Yes, I'm aware. My main concern is getting rid of _memcached_.
I might have a look at the sqlite stuff then. It's probably super easy.

renekliment:
Currently, there is no clear idea how (if) we're gonna use sqlite for storing stuff like timers, etc, so let's not go there now. I don't think it's a problem, getting the token each time AlexaPi restarts. Let's just get rid of memcached and we can do this when we know how we want to do this.

lechat:
Why not just keep that in a text file somewhere? OS provides timestamping and we don't need no SQL to find that file.

renekliment:
[...]

Issue #23, opened 2016-09-24
Refactor: AVS


renekliment:
Right now, there's a lot of intertwined code regarding AVS communication (requests, response handling), triggering, recording, playback.

Splitting AVS into a separate module is especially important since we could then potentially support other services if available (Google Assistant for example), but what's more important, we could then start the work on new AVS API without much problems.

renekliment:
Note: Have a look at @ewenchou's https://github.com/ewenchou/alexa-client if it makes any sense to use it.

PR #138, opened 2017-01-07
Get rid of memcached & store the token in runtime memory

renekliment:
**Issue this PR fixes (if applicable):** #43

**Description:** What the title says. When doing AVS code refactor (#23), the `Token` will of course be moved to a module.

**Checklist:**
  - [x] Update CHANGELOG.md
  - [x] Update Wiki (new platform, installation, configuration changes etc)
  - [x] rebase after the logging stuff is merged


src/main.py
```diff
@@ -233,26 +227,47 @@ def internet_on():
...
+	def __init__(self, aconfig):
+
+		self._aconfig = aconfig
+
+		self.renew()
+
+	def get(self):
+
+		if time.time() - self._timestamp > self._validity:
+			self.renew()
```
lechat: Perhaps it is worth adding a debug log here stating that token is expired.

src/main.py
```diff
@@ -233,26 +227,47 @@ def internet_on():
...
+
+		return self._token
+
+	def renew(self):
+
+		logger.debug("AVS token: Requesting a new one")
```
lechat: And this should be `info`, this is good thing to see in log before enabling verbose logging.

src/main.py
```diff
@@ -233,26 +227,47 @@ def internet_on():
...
 		resp = json.loads(response.text)
-		mc.set("access_token", resp['access_token'], 3570)
-		return resp['access_token']
-	else:
-		return False
 
+		self._token = resp['access_token']
+		self._timestamp = time.time()
+
+		logger.debug("AVS token: Obtained successfully")
```
lechat: This could be `info` too - it is good to know that we've asked for token and obtained it.

src/main.py
renekliment: When the token expires, you get a renew message. So you'd see an info message and a debug message? Maybe we should change the one message to be more informative?

src/main.py
```diff
@@ -233,26 +227,47 @@ def internet_on():
...
+	_timestamp = None
+	_validity = 3570
+
+	def __init__(self, aconfig):
+
+		self._aconfig = aconfig
+
+		self.renew()
+
+	def get(self):
```
lechat: Just as idea: rename this method to `token` and decorate it with `@property', then call to this will look like: 

```
headers = {'Authorization': 'Bearer %s' % token.token}
```
or even changing that method to `__str__`, so call above will look like this:
```
headers = {'Authorization': 'Bearer %s' % str(token)}
```

src/main.py
```diff
@@ -233,26 +227,47 @@ def internet_on():
...
+			self.renew()
+
+		return self._token
+
+	def renew(self):
+
+		logger.debug("AVS token: Requesting a new one")
+
+		if not self._aconfig['refresh_token']:
+			raise ValueError("refresh_token is not valid")
```
lechat: This check and, especially, exception must happen in `__init__`. If it is not done there, then lines 279 and 320 may break and it will be harder to debug that case.

src/main.py
lechat: True, but then that unified message will be like "Token does not exist or expired". As if we don't really know. I personally hate such messages and prefer them to be explicit.

renekliment:
Rebased. Addressed the logging comments, thank you. I didn't add another commit, since it's such a small change. I hope it's okay.

renekliment:
Check moved, good point. Let me know if you have further comments or I can merge this. Thank you.

src/main.py
```diff
@@ -263,6 +263,9 @@ def __init__(self, aconfig):
 
 		self._aconfig = aconfig
 
+		if not self._aconfig['refresh_token']:
```
lechat: If `_aconfig` does not have that key, how does this prevents `Token` to progress? I think it is better to log warning, add that key and go on.

renekliment: Well, the exception crashes AlexaPi.

We can't go on. The setup (auth_web.py) adds refresh_token and we need it in order to obtain AVS token to communicate.

lechat: Ah, sorry, that token is something different to what I thougt. I think that it is better to change message describing what we were after and what needs to be added to config or what to run to re-add it (I guess this is web_auth?)

Having messages that describe possible fix will reduce support load on us in future if users can find and fix errors themselves.

renekliment: Okay. How about 

```python
logger.critical("AVS refresh_token not found in the configuration file. Run the setup again to fix your installation (see project wiki for installation instructions).")
sys.exit()
```

I could say the auth_web script instead of setup, but it's probable that when the token is missing, setup wasn't run at all or got messed up. Is that okay?

lechat: `sys.exit(1)` - this is error condition, so we exit with non-zero code.

lechat: And message it good!

renekliment: Yeah, that's what I meant. I thought I'd check before commiting it - too late :smile:

renekliment:
Okay to merge?

renekliment:
Hmm, I should probably do a mention ... @lechat

src/main.py
```diff
@@ -257,26 +252,50 @@ def internet_on():
...
-			"refresh_token": refresh,
+			"client_id": self._aconfig['Client_ID'],
+			"client_secret": self._aconfig['Client_Secret'],
+			"refresh_token": self._aconfig['refresh_token'],
 			"grant_type": "refresh_token"
 		}
+
 		url = "https://api.amazon.com/auth/o2/token"
 		response = requests.post(url, data=payload)
 		resp = json.loads(response.text)
```
lechat: I think that it would be good to add here check for response error. Just in case that something is wrong, it is better to deal with human readable error rather than with python stack trace saying that json cannot be parsed.

renekliment: Addressed.

renekliment:
Rebased. Are there any more remarks? 
I'd like really to get rid of the memcached dependency - it makes me sad :disappointed:

lechat:
Дщщлы... oops! ;)

Looks good!

renekliment:
Thanks! Merging.

PR #142, opened 2017-01-10
Removed unnecessary use of global keyword

lechat:
**Description:**
This is just a small cleanup.

"global" keyword only required if function is going to change value of global variable. For read-only access or for changing property of a global class it is not required.

**Checklist:**
  - [x] PR Location `dev`


renekliment:
Haha, I know and I've put it there intentionally, because I want to get rid of the global mess eventually and this is a good indicator what we need to solve :smile:

renekliment:
Merged.

PR #144, opened 2017-01-14
Remove small duplicities

renekliment:
**Description:**
Remove small pieces of duplicate code.

**Checklist:**


src/main.py
```diff
@@ -415,7 +400,6 @@ def process_response(response):
 			for directive in j['messageBody']['directives']:
 				if directive['namespace'] == 'SpeechSynthesizer':
 					if directive['name'] == 'speak':
-						platform.indicate_recording(False)
```
enclarify: it might be helpful for review if this one line change had more explanation in the git commit message and possibly resided in it's own commit. It looks like this code line was removed in 82a97187186b2e203779ae719b0c076099d06207 and then came back in 2374632e0da13cc480aaf3887e970d5dca597573 (likely because of a rebase from another branch). This is not uncommon and  understandable but hard to review without a little more info

renekliment: It was removed in 852a572 cleanup and due to an incorrect rebase appeared in the mentioned commit. The logging stuff is also due to an incorrect rebase.
I'll make the change when I have the time. Thanks.

lechat:
Looks good!

renekliment:
@lechat Thanks!

@enclarify Split into two commits and they now have a better commit message. Appreciate your comments, I'll try doing better commits in the future. That should also help when there is a need to revert something.

Okay to merge now?

ImEmJay:
Looks good

renekliment:
Thanks, merging.

PR #146, opened 2017-01-14
Make internet_on() behave

renekliment:
**Description:**
Make it look and work better.

**Checklist:**


src/main.py
```diff
@@ -247,12 +247,11 @@ def mrl_fix(url):
...
 
 def internet_on():
-	logger.info("Checking Internet Connection...")
 	try:
 		requests.get('https://api.amazon.com/auth/o2/token')
 		logger.info("Connection OK")
 		return True
-	except:  # pylint: disable=bare-except
+	except requests.exceptions.RequestException:
 		logger.error("Connection Failed")
```
lechat: Would be good to say what exactly happened here, so it will be easier for end users to debug and they won't have to open support issue.

src/main.py
```diff
@@ -606,8 +605,9 @@ def cleanup(signal, frame):   # pylint: disable=redefined-outer-name,unused-argu
 	pHandler.setup()
 	platform.setup()
 
+	logger.info("Checking Internet Connection ...")
 	while not internet_on():
```
lechat: Why do we have that loop here? We're no retrying connection, `internet_on` is blocking and returns either `True` or `False`.

renekliment: Because we want the main part of AlexaPi to start after there is internet connection. It can take a while after the system start that the network is available. The systemd unit params should help, but this is a clear check.

lechat: So if we encounter unresolvable condition, like `ProxyError`, we will try again? I think that we should be specific on which errors we should retry and others, where retrying is pointless.

renekliment:
So I gave this another thought and I think we should start AlexaPi independently of the internet connection and handle runtime internet outages. Why:
- Internet connection outages can happen. There is no reason AlexaPi should crash. It should just work after the connection is resumed.
- If the connection is not available during startup, it should be the same as runtime. Crash / holdup on startup makes no sense.
- We will support alarms with the new API and we have to start AlexaPi fully to be able to do them on time even without the connection.

@alexa-pi/team Thoughts?

dgonano:
I absolutely agree. Mine was crashing sometimes with dodgy internet but it shouldn't.

lechat:
I agree. It should work and don't crash/exit if internet is not available. Report and continue.

renekliment:
Great! I've filed #159 and milestoned it to _v1.5_ because it's gonna take a while to do it properly and we should roll out _v1.4_ as soon as we can.

I'd like to still merge this for the meantime:
- It doesn't duplicate the _Checking_ message.
- If there is no connection, it patiently waits one second between re-attempt instead of running crazy fast loop to max out the CPU.

@lechat Is that okay?

Thank you.

renekliment:
We will do the offline mode properly later as discussed. I'm merging this in the meantime, since `internet_on()` goes crazy now on no internet connection.

PR #147, opened 2017-01-14
Fix future pylint messages

renekliment:
**Description:**
Fixes new pylint warnings introduced in pylint 2.0, namely
```Do not use `len(SEQUENCE)` as condition value (len-as-condition)```
and 
```Unnecessary "else" after "return" (no-else-return)```

This should wait after all currently open PRs get merged so it doesn't create merge conflicts.

**Checklist:**

### CODE REVIEW

File: src/alexapi/triggers/platformtrigger.py [hunk 1]
```diff
@@ -54,11 +54,11 @@ def platform_callback(self, platform_continuous_callback=None):
 				long_press_thread.start()
 
 	def continuous_callback(self):
-		if self._platform_continuous_callback:
-			return self._platform_continuous_callback()
-		else:
+		if not self._platform_continuous_callback:
 			return False
 
+		return self._platform_continuous_callback()
```
lechat: Could be a one liner:

```
return self._platform_continuous_callback() if self._platform_continuous_callback() else False
```
From another perspective, why do we bother keeping `self._platform_continuous_callback` variable and not let platform decide what to do about it?

renekliment: Because the relation is strictly this: `platform <---> trigger <---> recording` and the platform shouldn't get in touch with recording. Also, we use the callback in `long_press()` in the trigger.

renekliment:

maso27:
Hey @renekliment I just want to thank you for keeping this going.  I know that my life has been too busy for more than just the odd "trying to help people out" moments, but I want you to know that you're not operating in a vacuum.
I hope to get more time to work on some testing etc., but for now, know that I for one appreciate your efforts.

PR #148, opened 2017-01-14
Automatic platforms (devices) list in setup

renekliment:
**Description:**
I wanted to originally do that with the directory structure change, but didn't think of a good & clean solution, so I didn't do that. I don't have a clean solution now, the code is a little ugly, but I've decided to finally do that.
Also, the device _other_ has its own file now, which makes more sense than silly `if`s.

The reasons for this:
- all the device platform related code is contained neatly in one file
- when adding a new platform, people don't have to touch the main setup script just to add description
- as a reason of ^, there will no merge conflicts in the main setup script

**Checklist:**


lechat:
Looks good!

renekliment:
Thanks! Merging.

PR #149, opened 2017-01-14
Change pocketsphinx's default threshold

renekliment:
**Description:**
This should improve trigger word recognition. Unfortunately values like `1e-20` or even `1e-15` caused too many FPs for me so I've settled with this value.

**Checklist:**
  - [x] Update CHANGELOG.md
  - [ ] Update Wiki (new platform, installation, configuration changes etc)


renekliment:
Sadly, it turns out that `1e-10` is fine for music, but when I talk, AlexaPi goes crazy. But maybe we should do it, because people have various hardware and it may work better for the majority and other can always change it back?

dgonano:
I use 1e-30 which works well for me. 
I'd say set it higher rather than low, that way people know it's working and can tune it rather than getting annoyed with it and not realising

lechat:
Can we leave some known good values coummented out in config? That way people who is not that much familiar with possible settings can play with known good values at first.

renekliment:
It's described in the wiki - https://github.com/alexa-pi/AlexaPi/wiki/Triggers and there's a link to it in the config template.

So do you think `1e-10` is a good default value?

dgonano:
Yeah its better. If people still have issues we can increase it again

renekliment:
Yeah, makes sense. Merging.

PR #151, opened 2017-01-14
device_platforms: don't look for triggers when the platform trigger is disabled

renekliment:
**Description:**
Tiny device_platform cleanup.
Also removes the "verbose" option for the dummy platform from the config, because it isn't used anywhere.

**Checklist:**


renekliment:
My 8th PR today! This has to be some sort of a record!

@alexa-pi/team Please have a look at these one-line PRs, it should be pretty quick. Thank you.

src/config.template.yaml
```diff
@@ -132,8 +132,6 @@ platforms:
     min_seconds_to_record: 3
 
   dummy:
-    # whether to print messages when methods are called
-    verbose: false
```
lechat: I know that `dummy` don't use config, but for consistency each platform must expect that it's config is a `dict`. If you remove "verbose", dummy will receive `str` instead of `dict`. Better to put here explicit empty `dict`:

```
dummy: {}
```

src/alexapi/device_platforms/orangepiplatform.py
```diff
@@ -22,9 +22,10 @@ def after_setup(self, trigger_callback=None):
 
 		self._trigger_callback = trigger_callback
 
-		thread = threading.Thread(target=self.wait_for_button, args=())
-		thread.daemon = True
-		thread.start()
+		if self._trigger_callback:
+			thread = threading.Thread(target=self.wait_for_button, args=())
```
lechat: `trigger_callback` passed, but not used. Is that intentionally?

renekliment: It's used (called) in the `rpilikeplatform.py`.

src/config.template.yaml
renekliment: Good observation. Thanks.

dgonano: This is taken care of because in the `baseplatform` it says, if dummy, update with dummy. 
If its empty it won't attempt to add it. I originally did it like that because I was having this issue with an empty `common` section

dgonano:
Looks fine to me but I'm unfamiliar with the desktop platform.

src/config.template.yaml
lechat: @dgonano That's exactly my point. Having per platform permutation of contract is bad. What `baseplatform` should do is validate and print warning that specific platform may be misconfigured. Or make sure that platform gets `dict` no matter what (but that would be harder to implement, I guess).

renekliment: > permutation of contract

What do you mean by this?

Well, if you look at the code, `self._pconfig` is _always_ a dict.

renekliment:

PR #152, opened 2017-01-15
Don't ask the user about autorestart in setup (disabled by default)

renekliment:
**Description:**
So people often check this option, but have invalid config, so AlexaPI crashes and keeps on restarting. This is not the case intended for the feature.

This removes the setup option and people who do want that can easily enable it. It's all described in the wiki, now in its own page - https://github.com/alexa-pi/AlexaPi/wiki/Restart-on-crashes
It removes only the question (and defaults to disabled), but leaves the underlying methods parameter, so it can be later used when doing non-interactive setup or something.

**Checklist:**
  - [x] Update CHANGELOG.md
  - [x] Update Wiki (new platform, installation, configuration changes etc)


dgonano:
I don't really see the point in doing this. At least they know their program is having isssues. 
If it only says hello once yo may be trying to say "Alexa" and get fustrated

renekliment:
Hmm ... that's true ...

It's not a standard daemon behavior though. The only point of this endless loop is to notify the user that the config (or something else) is wrong. Also, it will do two requests to Amazon every time it starts (one to check the internet connection and one to obtain the access token, which we will not store anywhere permanently with #138 and that will be requested every time AlexaPi starts up).

lechat:
I like this change. When I first installed AlexaPi, I was puzzled by this question in setup: "Does that mean that program is so unstable, so it needs that special nudge to be stable enough? Or what? I better not enable this."
I think that it is better to leave such option for people who has crashing problem, hoping that this is small population.

maso27:
My opinion is that the auto restart option should be removed.
AlexaPi has advanced _significantly_ since I put that in there, and if there are instabilities with the code it can be addressed from a different angle than by restarting it with brute force.

renekliment:
Hm. So if we take this as a vote, we'll merge this.

Also, @dgonano you have a point that when you hear "hello", you expect that everything works. So, I've made a few changes:
- it checks if the `input_device` is valid (which is the main problem why AlexaPi crashes) and prints valid devices (this is useful, I don't know why we weren't doing this a long time ago)
- it plays the _hello_ only after all synchronous setup operations are finished (which also is checking the `input_device`)

It looks like this:
```
renekl@piggy ~/workspace/AlexaPi (git)-[disable-autorestart-setup] % python2 src/main.py -d                                    
2017-01-18 17:17:12 CRITICAL: Your input_device 'defaulta' is invalid. Use one of the following:
null
pulse
default
sysdefault:CARD=MID
front:CARD=MID,DEV=0
```

Is this okay?

renekliment:
Damn, the list isn't exhaustive, because you can address device as `hw:x,y` ... Okay, I'll rewrite it to make an attempt to open the device and if it doesn't work, let's just throw the damn exception with a meaningful message. We can then rewrite the mess when we use sounddevice instead of pyalsaaudio.

dgonano:
Yeah that last looks like a better approach to me :)

lechat:
While you can address device by "hw:x,y", is that what comes to mind of person, who just installed AlexaPi to try it out? I think that result from `alsaaudio.pcms(alsaaudio.PCM_CAPTURE)` is what alsaudio recognizes as valid capture devices and we shall stick to those.

At the same time it would be good to provide override option in case user knows that the only way that works is to use "hw:x,y". In config with message that dragons are expected there. ;)

renekliment:
Ah, silly me! I thought we have a problem when we set device for CHIP like this, but we actually set the `output_device`, so no problem there. 
We do have `input_device: "plughw:1"` in the config template though. I believe we should set it to `default` and let the people set something else if they want to.

Okay, let's try your idea @lechat :-) Sound good to me. I'll do that soon.

renekliment:
I've moved the `input_device` validation to #156, because it makes more sense.

This then:
- doesn't do restart by default
- indicates setup success (platform indication, playing hello) after all synchronous operations

dgonano:
I think I'm probably out of date on this one but from looking at the file changes it looks simple enough to approve.

renekliment:
Dope! Merging.

Issue #53, opened 2016-10-21
Snowboy support


renekliment:
We can see progress in this repo, so I'm opening this issue already, even though it's gonna take a while to get there.

After #24 is done, let's also support snowboy (so users can choose, let's not just replace pocketsphinx with it).

@maso27 and @flooie have done this in their repos based on the original project. We can learn from that.

renekliment:
Ah, it seems that snowboy isn't open-source or free (as in freedom). Just stating though, it's not a blocker. Good thing we keep support for pocketsphinx, so people who want to run free software only can run AlexaPi. But those people wouldn't want to use such _evil_ services like AVS. Well, nevermind :smile:

renekliment:
A couple of notes:
- I'd do it as a git submodule - there is no pypi package.
- example config:
```yaml
triggers:
  snowboy:
    enabled: false
    voice_confirm: true

    # rpi, ubuntu64, osx, android-armv7a
    platform: "rpi"

    model: "{distribution}/Alexa.umdl"
```

dgonano:
are you working on this @renekliment ? If so, how far away do you think it is?

my hotword detection isn't very reliable with sphinx and i'm hoping snowboy will improve it.. :p

renekliment:
No. It should be fairly easy to do though.

dgonano:
side note: store the personal model in /etc/opt/AlexaPi so it's kept during reinstalls

dgonano:
@renekliment i'm not sure where they are in the other alexapi repo's but we should be able to adapt their example and wrap it as a trigger yeah? 
https://github.com/Kitt-AI/snowboy/tree/master/examples/Python

If thats what your were thinking i'll have a look at it tomorrow

renekliment:
@dgonano Yes. I said it should be fairly easy :smile: We just need to use _sounddevice_ instead of _pyaudio_.

renekliment:
Hey, just checking in. Any progress on this?

dgonano:
No, I haven't had time to look at it and I'm going away this weekend for a week also.

renekliment:
@dgonano I think you've mentioned that snowboy works better for you. Have you just tested the example or do you have some preliminary AlexaPi implementation that you could share so others could base on top of that if you're too busy?

dgonano:
I think that was from the example. I haven't done any work on snowboy for this repo. Once I tuned sphinx mine was working fine so it hasn't been top priority.

[...]

PR #156, opened 2017-01-22
Simple Capture module

renekliment:
**Description:**
This: 
- moves the `silence_listener()` to an extra module to clean up a bit
- sets the default input device to `default`, which makes more sense
- validates the `input_device` config option and exits AlexaPi with a list of valid devices if the configured device is invalid
- provides an override to ignore a failed validation

**Notes:**
- This replaces some commits from #152.
- The voice triggers (pocketsphinx, snowboy) could in theory use some methods of alexapi.capture and be library-independent. They would only state things like framerate, format, periodsize and stuff to get an audio_input handle they could then just read from and close it. We can bother with this later after we have #45 and maybe #53 - we'll see how convenient that would be. 

@alexa-pi/team Please have a look. This should be a simple thing to review. Thanks!

**Checklist:**
  - [x] Update CHANGELOG.md
  - [ ] Update Wiki (configuration changes)


renekliment:
14 day rule ... merging.

PR #160, opened 2017-01-25
Fix unwanted variable expansion

renekliment:
**Description:**
- We should probably fix this in `master`, but meh, who cares, it's not a major issue. I still hope to roll out _v1.4_ soon.
- found by @maso27
- http://stackoverflow.com/questions/27920806/how-to-avoid-heredoc-expanding-variables


lechat:
I don't think that this is major issue that have to be fixed in master. So far none of the users stumbled on it.

renekliment:
Yep. It's a major problem in `dev`, since we have that strict mode that exists the script on non-existent variable.

renekliment:
@alexa-pi/team Btw, if you guys feel that any of the open PRs can be merged, don't be shy to share :smile:

dgonano:
im cherry picking all the easy PRs so some can be merged lol

renekliment:
Thanks! In `dev` with it!

PR #173, opened 2017-02-24
Docs update with v1.4

renekliment:
**Description:**
Small docs updates.

**Checklist:**


renekliment:
The build for `dev` and therefore this has failed because of https://github.com/larsimmisch/pyalsaaudio/issues/29 The PR build was fine, because it was built earlier with the older version of _pyalsaaudio_. If there is no fix to _pyalsaaudio_ before our version release, I'll push a version lock for the older version and I'm keeping it like this for the meantime.

renekliment:
Merging ... doing _v1.4_ release - yay!

PR #176, opened 2017-02-25
Change to use unique device serial number

sammachin:
I've made a change to the auth process so that instead of using `001` as the device serial number for all Alexa Pi's it now uses a unique ID based of the mac address.
This means that when you have multiple AlexaPi's all authed against the same amazon account they show up separately in the alexa app (http://alexa.amazon.com/spa/index.html#settings) eg `Sam's Alexa Pi` `Sams 2nd Alexa Pi` etc Amazon take care of auto device nameing. 
* This makes it easier for debugging as you can see which device did what,
* Its easier to revoke a single device
* Its more in line with the Amazon specs
* When we move to the V2 api I suspect it will be needed anyway for things like pushing the control of music events to the right device.

I belive `uuid.getnode` should be generic enough to work on all platforms and ~doesn't directly expose the mac address to help with privacy but~ should generally stay the same across re-installs on the same HW.


renekliment:
Dope! Seems simple enough.

- Please fix the build.
- Don't `import getnode as get_mac`, just use the original name. Especially since it is not the MAC address.
- Update `CHANGELOG.md` to reflect this
- I'll test & merge afterwards. 

Thank you.

sammachin:
I've fixed the get_mac issue, sorry that was me being lazy doing a copy paste from Stack Overflow ;)
Added to changelog.

I've amended my original comment as getnode does expose a number that trivial to reverse to a MAC address although this could be any mac address on the box or even a random number, personally I'm not really bothered if my MAC is shared with amazon but if anyone really wants to change this it could be hashed.

src/auth_web.py
```diff
@@ -7,6 +7,7 @@
 import yaml
 import cherrypy
 import requests
+from uuid import getnode
```
renekliment: Put this on top to fix the Travis CI build.

Also, 
```python
import uuid
...
uuid.getnode()
```

seems more readable.

src/auth_web.py
```diff
@@ -7,6 +7,7 @@
 import yaml
 import cherrypy
 import requests
+import uuid
```
renekliment: > Put this on top to fix the Travis CI build.

Imports from the standard library have to appear before other imports. This will fail the build again

```python
import uuid

import yaml
```

sammachin:
ah ok makes sense now I was just trying to work out what you meant by
putting it at the top

renekliment:
@sammachin https://travis-ci.org/alexa-pi/AlexaPi/jobs/205289858#L1087

sammachin:
OK I'm lost, why is pylint failing this build?

renekliment:
@sammachin If you follow that link, you will see:

```
************* Module auth_web
I: 50, 0: Locally disabling unused-argument (W0613) (locally-disabled)
C:  7, 0: Trailing whitespace (trailing-whitespace)
```

with the second line highlighted meaning there is trailing whitespace on line 7 in auth_web that you need to get rid of.

sammachin:
yay got there eventually!

renekliment:
Thanks, the PR is fine now.

I'm thinking again about the privacy side of this. I mean people share this stuff (MAC address) 99% of the time - they have their Wi-Fi on, they don't have Privacy Extensions enabled on IPv6, etc ... but what about the remaining privacy-aware 1% of the population ...
It might sound funny, using Alexa at all in this case, but I think AlexaPi is also about using Alexa with as much privacy as possible (since you truly have full control of when it's listening, etc).

I think the `auth_web.py` should ask the user about the serial ID - have this `uuid.getnode()` as default (simply Enter) and any user input as an alternative.

@alexa-pi/team Thoughts?

maso27:
I agree that offering an alternative input is a good idea.  I definitely have friends who are interested in AlexaPi for the privacy aspect.

sammachin:
How about hashing the value, this saves adding another step to the setup script and then having to store that value in the config file so that it can be picked up but auth_web etc.
I could just change the line to 
`"deviceSerialNumber": hashlib.sha256(str(uuid.getnode())).hexdigest()`

If someone is really that concerned about privacy and they want to use a totally random string that they change every time they re-install AlexaPi then they can edit the code themselves, but this solution should be great for 99.999% of users.

renekliment:
@sammachin That would help in one way, but if you had the hash could still verify if it corresponds to a specific MAC address or a list of them.

I'm postponing this to _v1.5_, which should come out soon, so we have more time to think about it / implement it / test it.

sammachin:
Been thinking a bit more about this I'd really like to get it closed off,

* I agree that hashing does still mean you can reverse the process using a rainbow table, I'm undecided if that makes hashing pointless or not. 

* The ID that is generated from the MAC could be one of several on the device there's no way to be sure which is used and/or if it isn't just a random number anyway.

* You are only sending this data to amazon not broadcasting it or making it publicly available, if you're that concerned about your privacy from Amazon then you ain't likely to be using Alexa in the first place!

* We need some sort of unique device ID, it makes having multiple devices on the alexa admin app very confusing.

* The ID really should persist across re-installs, otherwise people will see duplicate devices in their dashboard which will cause them to ask us questions, although the getnode() isn't perfect I think its good enough for a persistent ID.

I don't like the idea of having ever more configurable options in the startup script, it raises the barrier to entry by putting people off, I still think if this issue is that sensitive to you and you want to use Alexa you'll go through the source code and make you own tweaks anyway.

renekliment:
@sammachin Alright, let's sha256 the thing and be done with it. I'll test later today.

renekliment:
@sammachin Also, please rebase on `dev` and update the CHANGELOG.md - put the entry in the _new_ Unreleased section. Thank you.

sammachin:
👍

sammachin:
replaced by #192

PR #192, opened 2017-03-08
Unique ID in Device Serial Number

sammachin:
New PR based on current dev, to replace #176


renekliment:
@sammachin I've made a small adjustment so it works under Python 3 also.

renekliment:
Tested, works great, merging.

PR #193, opened 2017-03-08
Fix the Hello.mp3 file

sammachin:
Modified the rate to 24kHz and changed to Mono to match other alexa sound files,

Small fix but the hello.mp3 we were using which I think dates back to my alexaweb project! was 16Khz stereo MP3, the other files along with the responses from AVS are 24Khz Mono, this file wasn't playing right on my Jabra speaker it was too fast, have also tested this on my regular Pi


renekliment:
@sammachin Works fine (not that I was expecting anything else). 

Looking at the file, it seems like there's almost two seconds ... [of silence](https://www.youtube.com/watch?v=4zLfCnGVeL4) after the _[hello](https://www.youtube.com/watch?v=YQHsXMglC9A)_ itself. I'm thinking, should we crop it? Not that it really matters, since almost all parts of the initialization happen before playing this and in threads, so the only thing it blocks is platform indication of success (blinking green LEDs and shit), which no one cares about, but if it's there like to keep the audio output open, so there is no cut/mess after the sound, I believe one second should be enough. Usually this is needed before the sound, not after it, before the sound card wakes up, so I'm not sure if this actually makes sense ... Thoughts?

sammachin:
Yeah I did wonder about trimming the audio down but then it's not really a
log file and it's used at startup rather than as part of an interaction so
I figured it was bette left slightly longer than risk clipping it on
unknown hardware, particularly thinking of things like Bluetooth speakers
where the connection may need to be setup.

I take your point the we could trim the end down if we wanted to but it's
minimal gain.

renekliment:
Haha, okay :smile: Merging and I'll do a release during the day.

PR #251, opened 2017-06-01
Make French compatible

EmerickH:
**Issue this PR fixes (if applicable):** 
`This patch prevent crashing on French systems (with accents like éàè)`

**Description:**
`Just change default encoding to Latin-1`

**Checklist:**
  - [ ] Update CHANGELOG.md
  - [x] ~~Update Wiki (new platform, installation, configuration changes etc)~~
  - [ ] PR Location (usually `master` for hotfixes, `dev` for general development)
  - [ ] Make working!


renekliment:
Ah, I totally missed that we don't set encoding for Python 2 ... We should use UTF-8 however. See https://www.python.org/dev/peps/pep-0263/

EmerickH:
@renekliment Yes but it's also not working with utf-8, I don't know why

renekliment:
Well, I'm afraid that setting anything else than utf-8 will break it for other people. We have to debug it and make utf-8 work for everybody.

EmerickH:
Yes, the problem with utf-8 was decoding 0xe9 (é)
`UnicodeDecodeError: 'utf8' codec can't decode byte 0xe9`
It seems to have a [solution](https://stackoverflow.com/questions/12468179/unicodedecodeerror-utf8-codec-cant-decode-byte-0x9c), but the error was in python files I think (on Windows)

EmerickH:
If we use AlexaPi with Python3, it's no more needed:
https://stackoverflow.com/a/28127538

renekliment:
@EmerickH AlexaPi supports both Python 2.7 & Python 3.x. You can use it under Python 3 so it works for you.

You have to see the log details ... 
```
11.50s$ pylint --rcfile=pylintrc --ignore=tunein.py src/auth_web.py src/main.py src/alexapi
************* Module main
C:  2, 0: Trailing whitespace (trailing-whitespace)
C: 30, 0: Trailing whitespace (trailing-whitespace)
C:  5, 0: standard import "import os" should be placed before "import importlib" (wrong-import-order)
C:  6, 0: standard import "import logging" should be placed before "import importlib" (wrong-import-order)
C:  7, 0: standard import "import tempfile" should be placed before "import importlib" (wrong-import-order)
C:  8, 0: standard import "import signal" should be placed before "import importlib" (wrong-import-order)
C:  9, 0: standard import "import shutil" should be placed before "import importlib" (wrong-import-order)
C: 10, 0: standard import "import time" should be placed before "import importlib" (wrong-import-order)
C: 11, 0: standard import "import sys" should be placed before "import importlib" (wrong-import-order)
C: 12, 0: standard import "import threading" should be placed before "import importlib" (wrong-import-order)
C: 13, 0: standard import "import json" should be placed before "import importlib" (wrong-import-order)
C: 14, 0: standard import "import optparse" should be placed before "import importlib" (wrong-import-order)
C: 16, 0: standard import "import subprocess" should be placed before "import importlib" (wrong-import-order)
-----------------------------------
Your code has been rated at 9.91/10
```

We need to support both versions. There have been discussions about this. I'm not sure which issue / PR.

EmerickH:
@renekliment Ok! On my Windows I can't use Python3 but I've created a new branch on my repo!

PR #262, opened 2017-06-15
Change from VLC to SoX by default

EmerickH:
**Issue this PR fixes (if applicable):** 
VLC is very buggy and there's a lot of issue with it!
It would be better to switch to SoX by default.

**Checklist:**
  - [ ] Be sure that everyone is okay to switch
  - [ ] Test if it's working
  - [ ] Update CHANGELOG.md
  - [ ] Update Wiki (new platform, installation, configuration changes etc)


renekliment:
How is VLC buggy?
Also, SoX doesn't support playlists yet well.
We shouldn't uninstall vlc. We should install that sox module though. That should be a separate commit then.

renekliment:
@EmerickH Forgot to tag.

EmerickH:
@renekliment I'm back!
So it was to solve "choppy audio problems" but if SoX can't do playlists it's problematic...
Maybe not now!
It would be better to find a solution to get VLC working...

I've just bought a WRT3200 on OpenWRT so I will try to get AlexaPi working on it!

Emerick

PR #263, opened 2017-06-17
New readme

EmerickH:
Just some changes in Readme to make it better.
We can use the Alexa badge: https://developer.amazon.com/public/solutions/alexa/alexa-voice-service/content/marketing-brand-guidelines#badgeusage

(click [here](https://github.com/EmerickH/AlexaPi/blob/dev-new-readme/README.md) to see the render)

**Checklist:**
  - [ ] Update CHANGELOG.md
  - [x] ~Update Wiki (new platform, installation, configuration changes etc)~
  - [ ] Clean commits
  - [ ] Move logo


EmerickH:
Hello @renekliment !
I'll do this but actually I'm not on my computer.
Also I've created a new audio setup wizard, you can find it here: https://emerickh.github.io/AlexaPi-config/

Emerick

renekliment:
@EmerickH 
`sudo cp /opt/AlexaPi/src/config.template.yaml /var/opt/AlexaPi/config.yaml`

1. The path is wrong, it's `/etc/opt/AlexaPi/config.yaml`.
2. People should **never** copy the file like this. They will just overwrite their config (AVS, device and stuff) created by the setup script.

Also, the guide doesn't cover the main problem, which is people using desktop OS with user session PA. People need to set up system-wide PA, since the user-session PA just eats the sound card and there is nothing else you can do. Or do some ALSA dmix magic, which is more difficult.

I'm afraid this will bring confusion.

Also I'm sorry I don't have time for this project anymore creatively. Setting up audio isn't easy for beginners, which generates the half of the issue log. Another problem is that people don't even bother to read the docs, which tries to help with this non-trivial issue. Eventually I got fed up with with those totally incomplete issues and issues where it was clear that people just didn't bother. I hope you will have better luck.

renekliment:
@EmerickH Hey. Do you have intentions & time to clean this up?

EmerickH:
Moved to #408

PR #370, opened 2018-02-01
adding sudo package

lnxbil:
If you want to install `shairport-sync`, you need to have `sudo` installed, otherwise, it'll fail.

### DIFF
--- a/src/scripts/inc/os/debian.sh
+++ b/src/scripts/inc/os/debian.sh
@@ -2,7 +2,7 @@
 
 function install_os {
     apt-get update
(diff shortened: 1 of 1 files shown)


EmerickH:
If it's for `shairport-sync`, you must add it into the `install_shairport-sync` function
https://github.com/alexa-pi/AlexaPi/blob/c8adfbe3f7e09b40c7d03be0e6925be88b7cbe61/src/scripts/inc/os/debian.sh#L12

renekliment:
1. We ask users using `sudo` to run the install script, so here is too late.
2. Installing shairport-sync probably never should have been a part of this project in the first place (maybe a separate script, or a wiki page, but not in the main install).

